PR #21, opened 2021-06-16
update long pressed UI on messages

chungwwei:
👋  Mux, I made an update to the long pressed UI on messages, accept it if you find it worthy. :)


lib/components/chat_panel.dart
```diff
@@ -54,35 +80,96 @@ class _ChatPanelWidgetState extends State<ChatPanelWidget> {
...
+                                        title: Text('${message.author}',
+                                            style: TextStyle(fontSize: 20)),
+                                        tileColor: Colors.blueAccent,
+                                        trailing: CloseButton(),
+                                      ),
+                                      ListTile(
+                                        leading: Icon(Icons.favorite,
+                                            color: Colors.deepPurpleAccent),
+                                        title:
+                                            Text('Following Since Jan 11 2024'),
```
kevmo314, line 108: Can this data be fetched from the API instead of hardcoded?

lib/components/chat_panel.dart
```diff
@@ -54,35 +80,96 @@ class _ChatPanelWidgetState extends State<ChatPanelWidget> {
...
-                                    final model = Provider.of<UserModel>(
-                                        context,
-                                        listen: false);
-                                    model.unban(
-                                        model.channels.first, message.author);
-                                    Navigator.pop(context);
-                                  }),
-                            ]),
-                          );
+                              shape: RoundedRectangleBorder(
```
kevmo314, line 83: Leave out custom shapes, the material defaults usually look the most consistent.

lib/components/chat_panel.dart
```diff
@@ -54,35 +80,96 @@ class _ChatPanelWidgetState extends State<ChatPanelWidget> {
...
+                                  borderRadius:
+                                      BorderRadius.all(Radius.circular(20.0))),
+                              child: Wrap(
```
kevmo314, line 86: What does `Wrap` do here?

lib/components/chat_panel.dart
```diff
@@ -54,35 +80,96 @@ class _ChatPanelWidgetState extends State<ChatPanelWidget> {
...
+                                children: [
+                                  ListView(
+                                    shrinkWrap: true,
+                                    children: [
+                                      ListTile(
+                                        shape: RoundedRectangleBorder(
+                                            borderRadius: new BorderRadius.only(
```
kevmo314, line 93: Omit `new`, it's not necessary in the latest versions of dart.

lib/components/chat_panel.dart
chungwwei: Looks like the data can be fetched from the API, Link: [https://dev.twitch.tv/docs/api/reference#get-users-follows](https://dev.twitch.tv/docs/api/reference#get-users-follows). But I had no experience with pulling data from an API with dart.

lib/components/chat_panel.dart
chungwwei: It's not useful in the layout. I was playing with Wrap and Column widgets, forgot to remove it. Will remove.

PR #25, opened 2021-06-23
update quick link creation and launch

juaoose:
Hi,

- Duplicate links are not allowed.
- Empty links are not allowed.
- Web URLs are opened using ChromeSafariBrowser (other schemes are launched with url_launcher)

### DIFF
--- a/lib/components/quick_links_bar.dart
+++ b/lib/components/quick_links_bar.dart
@@ -17,13 +17,13 @@ class QuickLinksBar extends StatelessWidget {
             icon: Text(source.icon,
                 style: TextStyle(fontSize: 24, fontFamily: "MaterialIcons")),
             tooltip: source.name,
-            onPressed: () async => launchLink(source));
+            onPressed: () => launchLink(source));
       }).toList());
     });
   }
 
   void launchLink(QuickLinkSource source) async {
-    final bool isWebUrl =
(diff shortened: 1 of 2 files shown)


lib/components/quick_links_bar.dart
```diff
@@ -14,10 +17,20 @@ class QuickLinksBar extends StatelessWidget {
             icon: Text(source.icon,
                 style: TextStyle(fontSize: 24, fontFamily: "MaterialIcons")),
             tooltip: source.name,
-            onPressed: () async {
-              await launch(source.url.toString());
-            });
+            onPressed: () async => launchLink(source));
```
kevmo314, line 20: `async` isn't necessary here, `() => launchLink(source)` should work.

lib/components/quick_links_bar.dart
```diff
@@ -14,10 +17,20 @@ class QuickLinksBar extends StatelessWidget {
...
       }).toList());
     });
   }
+
+  void launchLink(QuickLinkSource source) async {
+    final bool isWebUrl =
```
kevmo314, line 26: `final isWebUrl`, the type isn't necessary since it can be inferred.

kevmo314 (approved):
Awesome, looks good just a few changes :)

PR #28, opened 2021-06-26
simplify config for functions

kevmo314:


juaoose:
Nice! Could one avoid two `firebase-admin` imports by doing `admin.ServiceAccount`? which is preferred?

juaoose (approved):

kevmo314:
> Nice! Could one avoid two `firebase-admin` imports by doing `admin.ServiceAccount`? which is preferred?

Good call, definitely preferred to have fewer imports :)

PR #32, opened 2021-07-01
add bot muting for tts

chungwwei:
For Code Review:

Added code for muting bots, but the toggle button on the setting page won't reflect its value on toggle change. The value of the toggle is changed, but the UI doesn't reflect that.


lib/models/tts.dart
```diff
@@ -42,4 +57,16 @@ class TtsModel {
       _tts.stop();
     }
   }
+
+  bool get isBotMuted {
+    return _isBotMuted;
+  }
+
+  set isBotMuted(bool value) {
```
kevmo314, line 65: The value doesn't update because `TtsModel` doesn't implement `ChangeNotifier`.

To keep it a bit simpler (because I totally did that as a shortcut and not really thinking ahead...), instead of extracting `TtsModel` use the `ChatHistoryModel` and put a getter/setter on it that sets the property on `TtsModel`.

Basically a property like `ttsEnabled`: https://github.com/muxfd/rtchat/blob/main/lib/models/chat_history.dart#L159

lib/models/chat_history.dart
```diff
@@ -144,4 +144,13 @@ class ChatHistoryModel extends ChangeNotifier {
...
   }
+
+  bool get ttsIsBotMuted => _ttsModule.isBotMuted;
+
+  set ttsIsBotMuted(bool value) {
+    _ttsModule.isBotMuted = value;
+    notifyListeners();
+  }
+
+  TtsModel get ttsModule => _ttsModule;
```
kevmo314, line 155: I don't think `ttsModule` needs to be exposed, remove the getter?

lib/models/tts.dart
```diff
@@ -42,4 +65,16 @@ class TtsModel {
...
+    _isBotMuted = value;
+    if (!value) {
```
kevmo314, line 75: We don't need these here, these lines stop any active/pending TTS when TTS gets turned off.

lib/models/tts.dart
```diff
@@ -21,10 +35,19 @@ class TtsModel {
     });
   }
 
+  String getMsgAuthor(String message) {
+    final parts = message.split(' ');
+    return parts[0];
+  }
+
   void speak(String message) {
```
kevmo314, line 43: Might be better to pass the author along with the message and construct the spoken message in `TtsModel`?

kevmo314 (approved):

PR #34, opened 2021-07-02
Update agent references to the firebase project id

juaoose:
This uses the service account to retrieve the project id and set it to the different pub/sub resources the agent uses.

I guess something similar could be done to some strings in the functions directory.

Note: I think using .env could be an alternative... but would require setting up `service_account.json` plus `.env` so maybe just using the json file is easier.

Also, can I remove the `build` script in `package.json`? Considering `run` uses ts-node and not the output of build.

### DIFF
--- a/agent/src/index.ts
+++ b/agent/src/index.ts
@@ -1,13 +1,17 @@
-import * as admin from "firebase-admin";
 import { Message, PubSub } from "@google-cloud/pubsub";
-import * as dotenv from "dotenv";
 import Bottleneck from "bottleneck";
+import * as dotenv from "dotenv";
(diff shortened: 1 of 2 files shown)


kevmo314:
I agree I think just using the json file is easier.

> Also, can I remove the build script in package.json? Considering run uses ts-node and not the output of build.

Yeah go ahead :) Actually the reason I added it was in hopes of adding CI that will verify that the code checked in builds without deploying, but that isn't written yet so it's not really needed.

kevmo314 (approved):

juaoose:
Well thats enough justification to keep `build`

PR #35, opened 2021-07-02
TTS adjustment speed/pitch

chungwwei:
For Code Review:

added sliders for TTS speed and pitch. The normal speed is 1 and the normal pitch is also 1. The range for speed and pitch are [0.1, 2] and [0.1, 3] respectively, but I'm not sure if the ranges are good enough for users.

Not implemented: maybe we can also have a sample text and a play button and a reset button to reset speed/pitch to 1.


kevmo314:
Sweet, change looks good. Agree on the preview button :)

kevmo314 (approved):

PR #51, opened 2021-07-16
fix: correctly run flutter analyze

kevmo314:


.github/workflows/flutter-analyze.yml
```diff
@@ -1,6 +1,13 @@
...
 
-on: push
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+  workflow_dispatch:
```
juaoose, line 10: Would removing the `:` for `workflow_dispatch` work? With it it feels like something else is expected after.

[Like this, not really sure about it.](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#example)

kevmo314: I don't think so, it's a weird artifact of yaml syntax that nodes with no children adjacent to nodes with children look funny. Removing the `:` is only allowed if `workflow_dispatch` is the only child.

See: https://www.techiediaries.com/github-actions-workflow-manual-http-post-triggers-workflow_dispatch-repository_dispatch/ or https://github.community/t/workflow-dispatch-event-not-working/128856

juaoose: ohhhh til

juaoose (approved):

Issue #47, opened 2021-07-16
[testing] write some tests


kevmo314:
We should probably start writing some tests...

The models are probably the easiest thing to test first. Getting some of the TTS logic tested would be nice.

Issue #23, opened 2021-06-16
[chat_history] tag highlighting


kevmo314:
Twitch bolds the `@muxfd` tags. Additionally, tagging the active user gets a red highlight and fancy style.

chungwwei:
For highlighting, we would like to match how Twitch implements its highlighting feature. It's light gray-ish when tagging other ppl in both light and dark modes. See pictures below.

![](https://cdn.discordapp.com/attachments/199396285734846465/866063335928627260/unknown.png)
![](https://cdn.discordapp.com/attachments/845327843659218944/857645700664786944/unknown.png)

PR #53, opened 2021-07-17
issues #47 & #23

re0078:
I'm not really experienced in flutter, and wanted to find a point to start with, so I added some simple tests to your project and a small change to make tags rich text in chat history. I might messed up some dependencies a little bit. Feel free to ignore all those changes :))


kevmo314:
Awesome! Could you attach a screenshot of your change?

@chungwwei worked on this code last so I've requested their review for the tag changes. I will take a look at the tests :)

Otherwise, excited to get your change in once we take a look, hopefully you'll get to see some streamers using it too.

test/channels_test.dart
```diff
@@ -0,0 +1,34 @@
...
+import 'package:rtchat/models/channels.dart';
+
+// This is the very first time I write a test in flutter. take it easy on me plz! :D
+void main() {
+  test('Create Channels Test', () {
+    var channel1 = Channel("provider", "channelId", "displayName");
+    var channel2 = Channel("provider", "channelId", "another_display_name");
+    var channel3 = Channel("another_provider", "channelId", "displayName");
+    var channel4 = Channel("provider", "another_channelId", "displayName");
+    assert(channel1 == channel2);
```
kevmo314, line 11: Instead of using `assert` in tests, use `expect(channel1, equals(channel2));` and so on below.

test/channels_test.dart
```diff
@@ -0,0 +1,34 @@
+import 'package:flutter_test/flutter_test.dart';
+import 'package:rtchat/models/channels.dart';
+
+// This is the very first time I write a test in flutter. take it easy on me plz! :D
+void main() {
+  test('Create Channels Test', () {
```
kevmo314, line 6: The string here should describe the test, so in this case something like

```
test('== compares provider and channelId', () {
   ...
});
```

and same for the test below.

test/channels_test.dart
```diff
@@ -0,0 +1,34 @@
...
+    var channel1 = Channel("provider", "channelId", "displayName");
```
kevmo314, line 7: Prefer `final` over `var` when the value stays constant.

kevmo314 (changes requested):
Great start on the tests! Left some comments, could you address the `flutter analyze` issues as well?

Let me know if you have any questions, we're happy to help.

lib/components/chat_history/twitch/message.dart
```diff
@@ -116,7 +116,10 @@ Iterable<InlineSpan> parseText(
         ),
       );
     } else {
-      return TextSpan(children: tokenize(element.text, tagStyle).toList());
+      return WidgetSpan(
+        child: RichText(
+            text: TextSpan(children: tokenize(element.text, tagStyle).toList())),
+      );
```
chungwwei, line 122: I would keep the old code. it seems like the change is introducing weird changes to the font size of a message. It looked like the picture below.
image

lib/components/chat_history/twitch/message.dart
```diff
@@ -187,7 +190,7 @@ class TwitchMessageWidget extends StatelessWidget {
       final tagStyle = Theme.of(context)
           .textTheme
           .bodyText2!
-          .copyWith(fontSize: styleModel.fontSize, fontWeight: FontWeight.bold);
+          .copyWith(fontSize: styleModel.fontSize, fontWeight: FontWeight.w900);
```
chungwwei, line 193: here, adding a grayish `backgroundColor` might just do the highlighting part.

PR #59, opened 2021-07-18
feat: add bttv/ffz emote support

juaoose:
This adds support for BTTV emotes for chat history.

- I am not conviced the current parsing is very efficient?
- I used the mechanism you used for badges, does it apply
- Would FFZ ones fit in the same model (renamed) or would it be better to have another ChangeNotifier for FFZ?


kevmo314:
> I am not conviced the current parsing is very efficient?
> I used the mechanism you used for badges, does it apply

Yes this is fine :)

> Would FFZ ones fit in the same model (renamed) or would it be better to have another ChangeNotifier for FFZ?

I think putting them in the same model is fine. We can call it ThirdPartyEmotesModel or something.

lib/components/chat_history/twitch/message.dart
```diff
@@ -277,4 +272,29 @@ class TwitchMessageWidget extends StatelessWidget {
...
+      final emote =
+          bttvEmotes.globalEmotes[token] ?? bttvEmotes.channelEmotes[token];
+      if (emote != null) {
+        final url = "https://cdn.betterttv.net/emote/${emote.id}/1x";
+        children.add(WidgetSpan(
+            alignment: PlaceholderAlignment.middle,
+            child:
+                Image(image: NetworkImage(url), height: styleModel.fontSize)));
+      } else {
+        children.addAll(parseText(
```
kevmo314, line 291: I think the biggest issue is that this creates one `Widget` per token which can be a lot of widgets on screen. We can improve this by iterating and manually finding tokens. Actually, this algorithm is not too different from the one @chungwwei wrote for tags although maybe a little more complicated :)

```
var lastParsedStart = 0;
for (var start = 0; start < message.length; ) {
  final end = message.indexOf(" ", start) + 1;
  final token = end == 0 ? message.substring(start) : message.substring(start, end);
  if (bttvEmotes. ... [token.trim()] != null) {
    // this token is an emote. parse off the text from the last parsed start.
    children.addAll(parseText(message.substring(lastParsedStart, start)), ...);
    // then add the widget for the emote.
    children.add(WidgetSpan(...));
    // then move the last parsed start pointer.
    lastParsedStart = end;
  } else {
    // this token is not an emote, so try the next token.
    start = end;
  }
}
if (lastParsedStart != message.length) {
  // add the tokens that are after the last emote.
  children.addAll(parseText(message.substring(lastParsedStart)), ...);
}
```

kevmo314 (approved):

lib/components/chat_history/twitch/message.dart
juaoose: Would this function make sense as a top level function just like the other parse functions in this file?

kevmo314: Eh I think it's fine to put in the top level if you want. It can be made private by prefixing the name with an underscore. Static functions and top-level functions are basically the same in dart, I think it's just up to preference.

kevmo314 (approved):
Code looks good :) I think we should refactor the processing into the model in the future bc there's some benefits to that for future features, but that can be done in a separate PR. I'll take a look after this one gets in.

Nice work!

juaoose:
> Code looks good :) I think we should refactor the processing into the model in the future bc there's some benefits to that for future features, but that can be done in a separate PR. I'll take a look after this one gets in.
> 
> Nice work!

I actually asked about the function location because I was thinking on how to test it compared to the other parsing functions in this file. And there is some refactoring to be made to integrate these to the emote picker too.

kevmo314:
> I actually asked about the function location because I was thinking on how to test it compared to the other parsing functions in this file. And there is some refactoring to be made to integrate these to the emote picker too.

Ah yeah, I think testing becomes easier if this complex logic is refactored into the model. But y'know, merge first think later. Lots of people will be very happy with BTTV support :)

PR #62, opened 2021-07-18
chore: upgrade kotlin, set backup attr

kevmo314:


juaoose (approved):
Wow I did not know about app backups, interesting!

PR #63, opened 2021-07-19
feat: subscription logging

kevmo314:
Adding the last subscription timestamp so we can prune off channels that nobody is listening to to reduce bandwidth and database writes.


juaoose:
The idea would be to periodically look at the subscribed channels time stamp to see if old and then disconnect? From the agent?

kevmo314:
> The idea would be to periodically look at the subscribed channels time stamp to see if it's "old" and then disconnect? From the agent?

Yes, we'll add a cronjob that will delete subscribed channels if they haven't been pinged in some amount of time, maybe a week or so.

juaoose (approved):

PR #66, opened 2021-07-20
feat: add loading state for sign in screen

kevmo314:


juaoose (approved):
Nice, it was a bit weird going back to the login screen for a few seconds

PR #69, opened 2021-07-20
chore: clean up scrollview

kevmo314:


lib/components/pinnable/scroll_view.dart
```diff
@@ -0,0 +1,102 @@
...
+        break;
+      }
+      final pinned = isPinnedBuilder(nextPinnableIndex)!;
+      final sliver = PinnableMessageSliver(
+        vsync: vsync,
+        pinned: pinned,
+        child: AnimatedContainer(
+          duration: const Duration(milliseconds: 300),
+          curve: Curves.easeOut,
+          color: pinned ? Theme.of(context).primaryColor : Colors.transparent,
```
juaoose, line 59: Not exactly related to the task you did here, but maybe use a different color considering light mode? If I understand correctly this would make the whole sliver blue for a bit?

juaoose (approved):

PR #71, opened 2021-07-22
fix: use dividerColor instead of primaryColor for pins

kevmo314:


juaoose:
Did you get my deleted comment? I deleted it as I was not sure it was right

kevmo314:
> Did you get my deleted comment? I deleted it as I was not sure it was right

Ohhh that's why it didn't show up on github haha. I think it's right though, we should match the background color of pinned elements to the dividers, not the blue primary color.

juaoose (approved):

PR #75, opened 2021-07-23
fix: disabled long press menu when viewing some other channel

chungwwei:


lib/components/chat_history/twitch/message_link_preview.dart
```diff
@@ -1,6 +1,8 @@
 import 'package:metadata_fetch/metadata_fetch.dart';
 import 'package:flutter/material.dart';
 
+Map<String, _TwitchClipData> linkPreviewCache = {};
```
kevmo314, line 4: Can you remove the link preview change from this PR?

lib/components/chat_history/message.dart
```diff
@@ -28,6 +28,14 @@ class ChatHistoryMessage extends StatelessWidget {
         if (layoutModel.isInteractionLockable && layoutModel.locked) {
           return child;
         }
+        final userModel = Provider.of<UserModel>(context, listen: false);
+        final loginChannel = userModel.userChannel!.displayName;
```
kevmo314, line 32: Can we compare by channel id? The format for `TwitchMessageModel.channelId` is `$provider:$channelId`. This is more robust when we add other providers.

lib/components/chat_history/twitch/message_link_preview.dart
chungwwei: I copied and pasted the file `message_link_preview.dart` from main to my local one, and pushed it.

lib/components/chat_history/message.dart
chungwwei: I added `channelId` to `TwitchMessageModel`, and I changed it so its compared by `channelId`.

kevmo314 (approved):

PR #76, opened 2021-07-28
fix: replace underscore with spaces

kevmo314:


juaoose (changes requested):
I believe this change needs to be applied in the `spokenMessage` getter too.

kevmo314:
> I believe this change needs to be applied in the `spokenMessage` getter too.

I thought about this some more, I think we should just replace all underscores so it catches stuff in the message as well. Updated the PR.

juaoose (approved):

PR #89, opened 2021-07-30
feat: make quick links overflow

kevmo314:


juaoose:
I think the gradle upgrade was added here on accident 😄 I disabled auto merge if you prefer it staying on the original PR

juaoose (approved):

kevmo314:
> I think the gradle upgrade was added here on accident 😄 I disabled auto merge if you prefer it staying on the original PR.
> 
> I guess the only possible downside is that it might not be obvious for users that there are more icons if you swipe.

Oops fixed the gradle thing.

I sort of agree it's not super obvious, maybe we should add a shadow hint or something to indicate that there's more icons,

PR #93, opened 2021-08-01
feat: always enable speaker disconnect protection

kevmo314:
I don't really know why you would ever _not_ want speaker disconnect protection, perhaps we should keep it always enabled?


juaoose (approved):
I guess this is only relevant for external speakers right? I guess it makes sense and there is no considerable "penalty" for when there is no speaker.

PR #97, opened 2021-08-01
feat: option to show deleted messages

kevmo314:
Also fixes the agent's deleted message logic.


agent/src/client.ts
```diff
@@ -38,18 +36,22 @@ export function buildClient() {
...
-        });
+      const messageId = tags["target-msg-id"];
+
+      const original = await messages.doc(`twitch:${messageId}`).get();
+
+      if (!original.exists) {
+        return;
+      }
+
+      await messages.doc(`twitch:-${messageId}`).set({
```
juaoose, line 47: Was this meant to be `twitch:${messageId}`?

juaoose (approved):

PR #122, opened 2021-08-05
feat: add follow event to chathistory

chungwwei:


lib/components/chat_history/message.dart
```diff
@@ -121,6 +122,8 @@ class ChatHistoryMessage extends StatelessWidget {
       return TwitchRaidEventWidget(m);
     } else if (m is StreamStateEventModel) {
       return StreamStateEventWidget(m);
+    } else if (m is TwitchFollowEventModel) {
+      return TwitchFollowEventWidget(m);
```
kevmo314, line 126: Can we render this only in debug mode for now?

```
return kDebugMode ? TwitchFollowEventWidget(m) : Container();
```

kevmo314 (approved):
Nice!

PR #123, opened 2021-08-06
feat: add sub events to chat history

juaoose:
This adds subscription events to the chat history.

Note: For now, the message sent with resubs is not handled.

[Light Mode](https://user-images.githubusercontent.com/994594/128444867-c98b18ab-087c-4731-a5d7-df3d87459344.jpg)
[Dark Mode](https://user-images.githubusercontent.com/994594/128447108-a18c823f-4cc9-45c6-bd28-05c18d8808db.png)

Side note: Would it make sense to randomize the color of the BorderSide to more easily differentiate each event? (probably from a defined set of colors that work with each theme)

### DIFF
--- a/lib/components/chat_history/twitch/subscription_event.dart
+++ b/lib/components/chat_history/twitch/subscription_event.dart
@@ -32,10 +32,7 @@ class TwitchSubscriptionEventWidget extends StatelessWidget {
         child: Padding(
           padding: const EdgeInsets.fromLTRB(12, 4, 16, 4),
           child: Row(children: [
-            ClipRRect(
-              borderRadius: BorderRadius.circular(styleModel.fontSize),
-              child: Icon(Icons.star, size: styleModel.fontSize * 1.5),
-            ),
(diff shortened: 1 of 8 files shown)


lib/components/chat_history/twitch/subscription_event.dart
```diff
@@ -0,0 +1,58 @@
...
+            left: BorderSide(
+              width: 4,
+              color: Theme.of(context).accentColor,
+            ),
+          ),
+        ),
+        child: Padding(
+          padding: const EdgeInsets.fromLTRB(12, 4, 16, 4),
+          child: Row(children: [
+            ClipRRect(
```
kevmo314, line 35: Is the `ClipRRect` necessary?

kevmo314 (approved):

lib/components/chat_history/twitch/subscription_event.dart
juaoose: It is useless, removed it.

PR #135, opened 2021-08-07
fix: light mode icon color

kevmo314:


juaoose (commented):
Should we format quick link icons too?

kevmo314:
> Should we format quick link icons too?

Good catch, done :)

juaoose (approved):

PR #138, opened 2021-08-09
feat: add bit event to chat history

chungwwei:
image


lib/models/messages/twitch/event.dart
```diff
@@ -23,3 +23,19 @@ class TwitchFollowEventModel extends MessageModel {
     required bool pinned,
   }) : super(messageId: messageId, pinned: pinned);
 }
+
+class TwitchCheerEventModel extends MessageModel {
+  final int bits;
+  final bool isAnonymous;
+  final String cheerMessage;
+  final String giverName;
```
kevmo314, line 31: If the cheer is anonymous, is `giverName` null? Seems like it might be easier to make `giverName` nullable and make `isAnonymous` a getter?

lib/models/chat_history.dart
```diff
@@ -207,6 +207,16 @@ Stream<DeltaEvent> _handleDocumentChange(
...
       break;
+    case "channel.cheer":
+      final model = TwitchCheerEventModel(
+          bits: data['event']['bits'],
+          isAnonymous: data['event']['is_anonymous'],
+          cheerMessage: data['event']['message'],
+          giverName: data['event']['user_name'],
+          messageId: change.doc.id,
+          pinned: false);
+      yield AppendDeltaEvent(model);
```
kevmo314, line 218: Guard behind the flag.

lib/components/chat_history/twitch/cheer_event.dart
```diff
@@ -0,0 +1,65 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:intl/intl.dart';
+import 'package:provider/provider.dart';
+import 'package:rtchat/models/messages/twitch/event.dart';
+import 'package:rtchat/models/style.dart';
+
+class TwitchCheerEventWidget extends StatelessWidget {
+  final TwitchCheerEventModel model;
+  final NumberFormat _formatter = NumberFormat.compactSimpleCurrency();
```
kevmo314, line 10: Format as a number instead of currency.

lib/components/chat_history/twitch/cheer_event.dart
```diff
@@ -0,0 +1,65 @@
...
+            //   child: Image(
+            //       image: NetworkImageWithRetry(model.profilePictureUrl),
+            //       height: styleModel.fontSize * 1.5),
+            // ),
+            const SizedBox(width: 12),
+            Expanded(
+              child: RichText(
+                text: TextSpan(
+                  children: [
+                    TextSpan(text: name, style: boldStyle),
```
kevmo314, line 49: Is there a cheer emote chain associated with the message from twitch? Seems like it would be better to render those emotes.

functions/tsconfig.dev.json
```diff
@@ -0,0 +1,5 @@
+{
```
kevmo314, line 1: What's this change? :)

lib/models/messages/twitch/event.dart
chungwwei: both `isAnonymous` and `giverName` are always supplied in Firestore. The name is present in the database even if `isAnonymous` is `true`

lib/models/chat_history.dart
chungwwei: I didn't get this. I don't see any flagging in `chat_history.dart`.

lib/models/messages/twitch/event.dart
kevmo314: Wow that's surprising that Twitch would expose that. Haha okay :)

lib/components/chat_history/twitch/cheer_event.dart
chungwwei: I removed the formatter, and left the `bits` as an `int`.

lib/models/chat_history.dart
kevmo314: Oh sorry I misread the filename, disregard.

functions/tsconfig.dev.json
chungwwei: removed

kevmo314 (approved):

PR #143, opened 2021-08-09
chore: reduce unnecessary firebase deploys

kevmo314:


.github/workflows/firebase-merge.yml
```diff
@@ -1,6 +1,12 @@
 name: Deploy to Firebase on merge
 on:
   push:
+    paths:
```
juaoose, line 4: Should we include `database.rules.json`?

kevmo314: Good catch, done.

juaoose (approved):

PR #162, opened 2021-08-11
feat: render custom channel point redemptions

juaoose:
This adds custom channel point redemptions to chat history.

Notes:
- Default rewards don't seem to be notified through EventSub, they probably are through [IRC](https://github.com/tmijs/tmi.js/blob/master/lib/client.js)


lib/models/chat_history.dart
```diff
@@ -223,6 +224,32 @@ Stream<DeltaEvent> _handleDocumentChange(
...
+      final status = data['event']['status'];
+      if (status == "fulfilled") {
+        final model = TwitchChannelPointRedemptionEventModel(
+            pinned: false,
+            messageId: change.doc.id,
+            redeemerUsername: data['event']['user_name'],
+            status: status,
+            rewardName: data['event']['reward']['title'],
+            rewardCost: data['event']['reward']['cost']);
+        yield AppendDeltaEvent(model);
```
kevmo314, line 250: Should this be an UpdateDeltaEvent and modify the previous event?

kevmo314 (commented):
Hmm weird about the standard ones not being there but that's okay for now

lib/models/chat_history.dart
juaoose: I don't think it would need to update the previous one for regular scenarios since I'm just grabbing fulfilled events, maybe if the redemption could change from fulfilled to canceled.

But we just need some data to really see how these two events behave.

juaoose: I think it [might](https://help.twitch.tv/s/article/channel-points-guide?language=en_US) also change depending on how each streamer has it set up. I think for example some of the ones cat6 might have (sprint, cuss out loud) might never end up being fulfilled. And then I'd say automated ones should end up being fulfilled (like Pelotoms)

lib/components/chat_history/twitch/channel_point_event.dart
```diff
@@ -0,0 +1,64 @@
...
+                ],
+              ),
+            ),
+          )
+        ]),
+      ),
+    );
+  }
+
+  IconData defineIcon() {
```
kevmo314, line 50: Make this a getter instead.

kevmo314 (approved):
Nice job with the tests!

PR #169, opened 2021-08-13
feat: render hype train events

### LINKED ISSUE #156: Render hype train events

juaoose:
Closes #156


juaoose:
I am not sure if the message id change I did is ok?

lib/models/chat_history.dart
```diff
@@ -222,6 +224,70 @@ Stream<DeltaEvent> _handleDocumentChange(
           messageId: change.doc.id,
           pinned: false);
       yield AppendDeltaEvent(model);
+      break;
+    case "channel.hype_train.begin":
+      final model = TwitchHypeTrainEventModel(
+          pinned: true,
+          messageId: "train${data['event']['id']}",
```
kevmo314, line 231: I'd move all these params `data['event']...` stuff to `TwitchHypeTrainEventModel` as getters. That way, you can do something like `messageId: model.id` and it'll stay consistent. We can probably do like `TwitchHypeTrainEventModel.fromFirestore(DocumentSnapshot<...>)` so this code looks like `final model = HypeTrainModel.fromDocumentChange(change.doc);`.

That would also allow us to write some tests 😉  Otherwise this looks good!

tbh we should probably do that for all of the models too. Would be nice to get tests here.

lib/models/chat_history.dart
```diff
@@ -222,6 +224,70 @@ Stream<DeltaEvent> _handleDocumentChange(
...
+          total: data['event']['total']);
+      yield AppendDeltaEvent(model);
+      break;
+    case "channel.hype_train.progress":
+      yield UpdateDeltaEvent("train${data['event']['id']}", (message) {
+        if (message is! TwitchHypeTrainEventModel) {
+          return message;
+        }
+        // Since you can receive duplicates and order is not guaranteed
+        final level = data['event']['level'];
```
kevmo314, line 244: Also worth moving this logic to the model, so we can do something like:

```
if (message is! TwitchHypeTrainEventModel) {
  return message;
}
return message.withProgress(change.doc);
```

where `withProgress` performs this logic and returns a new model if necessary.

lib/models/chat_history.dart
juaoose: Would the best way to test this be testing fractions of chat_history model?

lib/models/chat_history.dart
```diff
@@ -264,12 +247,10 @@ Stream<DeltaEvent> _handleDocumentChange(
         if (message is! TwitchHypeTrainEventModel) {
           return message;
         }
-        return TwitchHypeTrainEndEventModel(
-            pinned: true,
-            messageId: "train${data['event']['id']}",
-            level: data['event']['level'],
-            total: data['event']['total'],
-            wasSuccessful: message.progress >= message.total ? true : false);
+        return TwitchHypeTrainEndEventModel.fromDocument(
```
kevmo314, line 250: Hmm this is kind of clever, I didn't really expect a use case where `UpdateDeltaEvent` would return a _different_ message model although I guess it works. Did you consider using the same model and rendering differently based on a getter? Might be less code but not sure.

Otherwise, looks good. :)

kevmo314 (approved):

lib/models/chat_history.dart
kevmo314: Nah, I'd write unit tests directly against the hype train model. Mostly verifying that the updating API works. We should one day simplify the chat_history model but it's not easy to test right now.

kevmo314:
> I am not sure if the message id change I did is ok? Maybe they could collide if I just used the event id?

Maybe a more appropriate prefix would be `channel.hype_train`? ie `channel.hype_train-${messageId}`. I'm not too concerned either way though, the event id's should have enough entropy to avoid collision. And worst case scenario we overwrite some historical events that are long gone anyways.

lib/models/chat_history.dart
juaoose: Well I also think it is kind of weird, but two visual items might be "too much"? Maybe something like updating the previous model to be hidden and then just append the end model could work? (Is deletion even possible?)

kevmo314: An `UpdateDeltaEvent` replaces the current event though, so it'll reuse the same widget. It shouldn't create a second visual item.

Deletion isn't possible right now, but in principle we could write a `DeleteDeltaEvent` and handle it too. I'm not sure it's necessary though...

lib/models/chat_history.dart
juaoose: Could it be better to not parse from `DocumentSnapshot` but from `Map<>`? So that the model doesnt have a dependency on the firebase package and that the tests dont require that sealed class?

kevmo314: Yeah that would be good too :)

kevmo314 (approved):

PR #170, opened 2021-08-13
fix: use different grey shade for dark/light pins

juaoose:


lib/components/pinnable/scroll_view.dart
```diff
@@ -29,6 +29,9 @@ class PinnableMessageScrollView extends ScrollView {
 
   @override
   List<Widget> buildSlivers(BuildContext context) {
+    final brightness = MediaQuery.of(context).platformBrightness;
```
kevmo314, line 32: Oh, does `Theme.of(context).brightness` work?

kevmo314 (approved):

lib/components/pinnable/scroll_view.dart
juaoose: Yes :D

PR #171, opened 2021-08-13
feat: added poll event to chat history

chungwwei:
image


lib/components/chat_history/twitch/poll_event.dart
```diff
@@ -0,0 +1,159 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:provider/provider.dart';
+import 'package:rtchat/models/messages/twitch/event.dart';
+import 'package:rtchat/models/style.dart';
+import 'dart:math' as math;
+
+List<Widget> getPollsWidget(List<dynamic> choices, List<Widget> children,
+    bool isCompleted, TextStyle baseStyle) {
+  num totalVotes = 0;
```
kevmo314, line 10: `int` instead of `num`

lib/components/chat_history/twitch/poll_event.dart
```diff
@@ -0,0 +1,159 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:provider/provider.dart';
+import 'package:rtchat/models/messages/twitch/event.dart';
+import 'package:rtchat/models/style.dart';
+import 'dart:math' as math;
+
+List<Widget> getPollsWidget(List<dynamic> choices, List<Widget> children,
```
kevmo314, line 8: It would be better to wrap this as a `StatelessWidget` instead of returning a list of widgets.

lib/components/chat_history/twitch/poll_event.dart
```diff
@@ -0,0 +1,159 @@
...
+    bool isCompleted, TextStyle baseStyle) {
+  num totalVotes = 0;
+  num totalChannelPointsVotes = 0;
+  num totalBitVotes = 0;
+  num maxVotes = 0;
+
+  for (final entry in choices) {
```
kevmo314, line 15: Can this logic be moved to the model? Same as in @juaoose's PR, if we move the computation stuff to the model it'll make it easier to test.

lib/components/chat_history/twitch/poll_event.dart
```diff
@@ -0,0 +1,159 @@
...
+    if (totalVotes > 0) {
+      percentage = votes / totalVotes;
+    }
+    if (isCompleted && maxVotes == votes) {
+      children.add(Padding(
+          padding: const EdgeInsets.fromLTRB(0, 0, 0, 8),
+          child: Stack(alignment: AlignmentDirectional.center, children: [
+            SizedBox(
+              height: 37,
+              child: LinearProgressIndicator(
```
kevmo314, line 40: Clever use of a progress indicator! :)

lib/components/chat_history/twitch/poll_event.dart
```diff
@@ -0,0 +1,159 @@
...
+    final votes = entry['votes'] ?? 0;
+
+    // avoid division by zero
+    double percentage = 0;
+    if (totalVotes > 0) {
+      percentage = votes / totalVotes;
+    }
+    if (isCompleted && maxVotes == votes) {
+      children.add(Padding(
+          padding: const EdgeInsets.fromLTRB(0, 0, 0, 8),
```
kevmo314, line 36: `EdgeInsets.only(bottom: 8)` here and same for the other `fromLTRB`'s with only one side.

lib/components/chat_history/twitch/poll_event.dart
```diff
@@ -0,0 +1,159 @@
...
+          child: Stack(alignment: AlignmentDirectional.center, children: [
+            SizedBox(
+              height: 37,
+              child: LinearProgressIndicator(
+                value: percentage,
+              ),
+            ),
+            Padding(
```
kevmo314, line 82: Can the conditional be put here? That way the rest of the widget doesn't have to be duplicated.

```
if (isCompleted && maxVotes == votes)
  const Padding(...)
else
  Container()
```

lib/models/chat_history.dart
```diff
@@ -223,6 +223,41 @@ Stream<DeltaEvent> _handleDocumentChange(
           pinned: false);
       yield AppendDeltaEvent(model);
       break;
+    case "channel.poll.begin":
+      final model = TwitchPollEventModel(
```
kevmo314, line 227: Move this to the model too so we can do like `TwitchPollEventModel.fromDocumentSnapshot(DocumentSnapshot<...>)` instead of parsing it all here.

kevmo314 (changes requested):

lib/components/chat_history/twitch/poll_indicator.dart
```diff
@@ -0,0 +1,80 @@
...
+                child: Align(
+                  child: Text(
+                    data.title,
+                    style: baseStyle,
+                  ),
+                  alignment: Alignment.centerLeft,
+                ),
+              )
+            ],
+            if (!isCompleted || data.votes < maxVotes) ...[
```
kevmo314, line 57: `else` instead of the inverted conditional.

Also the spread operator can be omitted since there's only one child.

lib/models/messages/twitch/event.dart
```diff
@@ -41,3 +42,75 @@ class TwitchCheerEventModel extends MessageModel {
...
+      {required this.id,
+      required this.title,
+      required this.bitVotes,
+      required this.channelPointVotes,
+      required this.votes,
+      required this.percentage});
+}
+
+class TwitchPollEventModel extends MessageModel {
+  final List<dynamic> choices;
```
kevmo314, line 63: Can we make this a `List<PollChoiceData>`? ie, parse in the constructor.

lib/models/messages/twitch/event.dart
```diff
@@ -41,3 +42,75 @@ class TwitchCheerEventModel extends MessageModel {
...
+    m.computeTotalVotes();
+    return m;
+  }
+
+  void computeTotalVotes() {
+    for (final entry in choices) {
+      final int votes = entry['votes'] ?? 0;
+      final int bitVotes = entry['bit_votes'] ?? 0;
+      final int channelPointVotes = entry['channel_point_votes'] ?? 0;
+      totalVotes = totalVotes + votes;
```
kevmo314, line 110: I would make these getters, a bit cleaner.

```
int get totalVotes => choices.reduce((sum, choice) => sum + choice.votes);
```

and such.

lib/components/chat_history/twitch/poll_indicator.dart
```diff
@@ -0,0 +1,80 @@
...
+  final int maxVotes;
+  const PollIndicatorWidget({
+    Key? key,
+    required this.data,
+    required this.isCompleted,
+    required this.maxVotes,
+  }) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
```
kevmo314, line 19: Instead of passing `percentage` through the model, calculate it here:

```
final percentage = data.votes / math.min(1, this.maxVotes);
```

lib/components/chat_history/twitch/poll_indicator.dart
```diff
@@ -0,0 +1,80 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:provider/provider.dart';
+import 'package:rtchat/models/messages/twitch/event.dart';
+import 'package:rtchat/models/style.dart';
+
+class PollIndicatorWidget extends StatelessWidget {
```
kevmo314, line 7: Call this `PollChoiceWidget` to match the model.

lib/models/messages/twitch/event.dart
```diff
@@ -41,3 +42,75 @@ class TwitchCheerEventModel extends MessageModel {
     required bool pinned,
   }) : super(messageId: messageId, pinned: pinned);
 }
+
+class PollChoiceData {
```
kevmo314, line 46: Call this `PollChoiceModel` to match the name of the other models.

kevmo314 (changes requested):
Almost there, looks great so far! A few more changes I think will help clean this up even more :)

lib/components/chat_history/twitch/poll_indicator.dart
chungwwei: I changed it to ```final percentage = data.votes / math.max(1, this.totalVotes);```

lib/components/chat_history/twitch/poll_indicator.dart
chungwwei: It didn't let me do the `else`. and `...[]` also seems necessary. IDE wasn't happy with either

lib/components/chat_history/twitch/poll_event.dart
```diff
@@ -0,0 +1,79 @@
...
+}
+
+class TwitchPollEventWidget extends StatelessWidget {
+  final TwitchPollEventModel model;
+
+  const TwitchPollEventWidget(this.model, {Key? key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return Consumer<StyleModel>(builder: (context, styleModel, child) {
```
kevmo314, line 28: You can use the new `DecoratedEventWidget` to simplify this now :)

kevmo314 (approved):
Nice work, looks good!

PR #182, opened 2021-08-16
fix: event-bits emote conditionals

chungwwei:


lib/components/chat_history/twitch/cheer_event.dart
```diff
@@ -5,9 +5,9 @@ import 'package:rtchat/models/messages/twitch/event.dart';
 import 'package:rtchat/models/style.dart';
 
 String getCorrespondingImageUrl(int bits) {
```
kevmo314, line 7: Can we add some unit tests for this? It also seems simpler to write something like:

```
final key = [10, 100, ...].firstWhere((k) => k < bits, orElse: () => 100000);
return 'https://.../$key/...';
```

chungwwei: added unit tests and made it simpler

kevmo314 (approved):
Nice :)

PR #191, opened 2021-08-18
chore: added eventsub_config model 

chungwwei:


lib/models/messages/twitch/eventsub_configuration.dart
```diff
@@ -0,0 +1,111 @@
+import 'package:flutter/material.dart';
+
+class EventSubConfigurationModel extends ChangeNotifier {
+  bool showFollowEvent = true;
```
juaoose, line 4: I think it would be better if the default config only enabled the less "spammy" events.

lib/models/messages/twitch/eventsub_configuration.dart
```diff
@@ -0,0 +1,111 @@
...
+  bool showCheerEvent = true;
+  bool showSubEvent = true;
+  bool showGiftedSubEvent = true;
+  bool showChannelPointEvent = true;
+  bool showHypeTrainEvent = true;
+  bool showPredictionEvent = true;
+  bool showPollEvent = true;
```
juaoose, line 11: Should we add raids here too? I think they follow the same concept as these events.

juaoose:
Also, something I am not sure about is if it would be better to split these settings at some point.

kevmo314:
> Also, something I am not sure about is if it would be better to split these settings at some point.

One thing that might be useful would be to create a separate non-`ChangeNotifier` object to encapsulate each of the groups that way the top level only includes the keys for each of the items. `AudioModel` does this: https://github.com/muxable/rtchat/blob/main/lib/models/audio.dart#L12

The equivalent might be doing something like `EventConfig` and then the `EventsubConfigModel` can be one `EventConfig` per type.

juaoose (approved):

Issue #11, opened 2021-06-10
[ai] message aggregation


kevmo314:
It's kind of annoying to get a million catJAM messages. Would be nice to aggregate emote-only or similar messages together into a more compact ui. ie, instead of

```
muxfd1: catJAM
muxfd2: catJAM
muxfd3: catJAM
```

We could do something like

```
muxfd1, muxfd2, +2 others: catJAM
```

This would need to be a little more intelligent around handling multiple emotes, but it would be really cool.

PR #193, opened 2021-08-20
feat: basic message compaction

kevmo314:
Screenshot_20210820-115057

First step for #11


juaoose:
The only thing that maybe we should consider asking for feedback is if the scrollable embedded chat examples might feel weird, since you cant scroll the page settings if you tap on that area (also since it might show an empty block when font size is small)

juaoose (approved):

kevmo314:
> The only thing that maybe we should consider asking for feedback is if the scrollable embedded chat examples might feel weird, since you cant scroll the page settings if you tap on that area (also since it might show an empty block when font size is small)

Yeah I noticed this too. I'll investigate how to improve that in a followup PR.

PR #194, opened 2021-08-21
Added print line to chat_history.dart

ryansng:
Print line checks if message received is prefixed by '!'. This is done through isCommand property of TwitchMessageModel.


lib/models/chat_history.dart
```diff
@@ -59,6 +59,10 @@ Stream<DeltaEvent> _handleDocumentChange(Map<String, List<Emote>> emotes,
           deleted: false,
           channelId: data['channelId']);
 
+      if (model.isCommand) {
```
kevmo314, line 62: This will trigger if anyone sends a command. Can we detect only outbound messages?

ryansng: Off my head I think it is possible to add a function `bool get isStreamer` to TwitchMessageModel. I will have to see how to implement the code though.

kevmo314: It would be simpler to instead check right before the user sends a message: https://github.com/muxable/rtchat/blob/a9e57b0415b2abe4cf4d6c725707163b54f4188d/lib/components/channel_panel.dart#L202

ryansng:
This PR is closed. Updates on https://github.com/muxable/rtchat/pull/200

PR #195, opened 2021-08-22
feat: added event configuration selection screen and follow screen

chungwwei:
- Events config selection screen added
- Only follow event config screen is added
- Toggle switches or slide bar do switch and slide, but the actual functions aren't implemented yet

image

image


lib/models/messages/twitch/eventsub_configuration.dart
```diff
@@ -1,9 +1,9 @@
 import 'package:flutter/material.dart';
 
 class FollowEventConfig {
-  final bool showEvent;
-  final bool isEventPinnable;
-  final int eventDuration;
+  bool showEvent;
+  bool isEventPinnable;
+  double eventDuration;
```
kevmo314, line 6: Dart also has a `Duration` type you can use here: https://api.dart.dev/stable/2.13.4/dart-core/Duration-class.html

lib/screens/settings/follow_event_config.dart
```diff
@@ -0,0 +1,59 @@
...
+                      min: 2,
+                      max: 14,
+                      divisions: 4,
+                      label: "${model.followEventConfig.eventDuration}seconds",
+                      onChanged: (value) {
+                        model.setFollowEventDuration(value);
+                      },
+                    ),
+                    SwitchListTile.adaptive(
+                      title: const Text('show event'),
```
juaoose, line 39: Should we follow the capitalization 'standard' that has been used in the settings? "Enable event"/"Show event"

lib/screens/settings/settings.dart
```diff
@@ -50,6 +50,13 @@ class SettingsScreen extends StatelessWidget {
               layoutModel.isStatsVisible = value;
             },
           ),
+          ListTile(
+            title: const Text('Event Configuration'),
+            // subtitle: const Text("Control which badges are visible"),
```
juaoose, line 55: Maybe we should have a simple subtitle to make it clear what event configuration is.

"Control how and which Twitch events are displayed" Names are hard but something like that 😅

lib/screens/settings/follow_event_config.dart
```diff
@@ -0,0 +1,59 @@
...
+                      value: model.followEventConfig.showEvent,
+                      onChanged: (value) {
+                        model.setFollowEventShowable(value);
+                      },
+                    ),
+                    SwitchListTile.adaptive(
+                      title: const Text('able to pin event'),
```
juaoose, line 46: Similarly, maybe we should explain what event pinning means with a subtitle?

lib/models/messages/twitch/eventsub_configuration.dart
chungwwei: changed the type from `double` to `Duration`.

lib/screens/settings/follow_event_config.dart
chungwwei: Agree.

lib/main.dart
```diff
@@ -235,6 +246,9 @@ class App extends StatelessWidget {
             '/settings/audio-sources': (context) => const AudioSourcesScreen(),
             '/settings/quick-links': (context) => const QuickLinksScreen(),
             '/settings/backup': (context) => const BackupScreen(),
+            '/settings/events-config': (context) => const EventConfigScreen(),
+            '/settings/follow-config': (context) =>
```
kevmo314, line 250: Since this is a child of `events-config` we should probably do `/settings/events/follow`

lib/main.dart
```diff
@@ -235,6 +246,9 @@ class App extends StatelessWidget {
             '/settings/audio-sources': (context) => const AudioSourcesScreen(),
             '/settings/quick-links': (context) => const QuickLinksScreen(),
             '/settings/backup': (context) => const BackupScreen(),
+            '/settings/events-config': (context) => const EventConfigScreen(),
```
kevmo314, line 249: `config` isn't necessary as it's under `/settings`. `/settings/events` is sufficient.

lib/screens/settings/eventsub_page.dart
```diff
@@ -0,0 +1,62 @@
+import 'package:flutter/material.dart';
+import 'package:provider/provider.dart';
+import 'package:rtchat/models/layout.dart';
+
+class EventConfigScreen extends StatelessWidget {
```
kevmo314, line 5: Call this `EventsScreen` and update the filename to match? `Config` isn't really necessary because it's under `settings`.

lib/screens/settings/follow_event_config.dart
```diff
@@ -0,0 +1,63 @@
+import 'package:flutter/material.dart';
+import 'package:provider/provider.dart';
+import 'package:rtchat/models/messages/twitch/eventsub_configuration.dart';
+
+class FollowEventConfigWidget extends StatelessWidget {
```
kevmo314, line 5: Same here. I'd also put these screens into a separate directory, eg `/settings/events/follow.dart`

lib/models/messages/twitch/eventsub_configuration.dart
```diff
@@ -1,76 +1,99 @@
...
-      SubscriptionEventConfig(false, false, 5);
-  CheerEventConfig cheerEventConfig = CheerEventConfig(true, true, 5);
-  // other event configs
+      SubscriptionEventConfig(false, false, const Duration(seconds: 5));
+  CheerEventConfig cheerEventConfig =
+      CheerEventConfig(true, true, const Duration(seconds: 5));
+  // other configs
   // final HypeTrainEventConfig;
 
+  setFollowEventDuration(int value) {
```
kevmo314, line 71: Accept `Duration` as the arg?

lib/screens/settings/settings.dart
```diff
@@ -50,6 +50,13 @@ class SettingsScreen extends StatelessWidget {
               layoutModel.isStatsVisible = value;
             },
           ),
+          ListTile(
```
kevmo314, line 53: Put this behind the flag so it doesn't get released just yet. :)

kevmo314 (approved):

Issue #61, opened 2021-07-18
[chat_history] add keyboard shortcut buttons for commonly used messages


kevmo314:
Perhaps we can detect when the user types a command like `!b` and add it to the list and keep track of the most used commands in an LRU cache.

PR #200, opened 2021-08-26
feat: add command shortcut bar (issue: #61)

ryansng:
Change log: 
1. User commands (i.e. prefixed by !) are cached in shared preferences
     - commands are cached automatically when user types and submits
2. Added command shortcut bar above keyboard
3. Fix shortcut bar not dismissing using flutter_keyboard_visibility package

To Fix:
 1. 'Bottom overflow' when layout and keyboard is opened (https://github.com/muxable/rtchat/issues/215)

Future to do:
1. (?) Toggle command suggestions
    - Enable / Disable
    - Always on / Only appear on keyboard open
    
Shortcut bar behaviour:
1. Bar will not display when there are no cache commands
2. Commands are sorted in order (most recently used on the left)
3. Least recently used command is deleted when the maximum number of cached command has exceeded
4. No limit set on maximum size of a single command

image


lib/components/channel_panel.dart
```diff
@@ -234,10 +240,59 @@ class _ChannelPanelWidgetState extends State<ChannelPanelWidget> {
...
+          },
+          style: TextButton.styleFrom(
+            primary: Colors.red,
+            textStyle: const TextStyle(fontWeight: FontWeight.bold),
+          )),
+    );
+    return commandButtons;
+  }
+
+  Widget _buildCommandsShortcutBar(CommandsModel commandsModel) =>
```
kevmo314, line 278: I think we should inline these functions. Splitting them up like this is makes the reading order not match the layout order. To reduce the nesting, make a `CommandButtonWidget` or similar. This also makes it easier to test if you so choose :)

lib/components/channel_panel.dart
```diff
@@ -234,10 +240,59 @@ class _ChannelPanelWidgetState extends State<ChannelPanelWidget> {
...
+      _chatInputFocusNode.hasFocus &&
+              commandsModel.commands
+                  .isNotEmpty //keyboard is opened and cache is not empty
+          ? SizedBox(
+              height: 55,
+              child: Scrollbar(
+                isAlwaysShown: false,
+                child: ListView(
+                  scrollDirection: Axis.horizontal,
+                  children: <Widget>[
```
kevmo314, line 288: `<Widget>` is not necessary

lib/components/channel_panel.dart
```diff
@@ -234,10 +240,59 @@ class _ChannelPanelWidgetState extends State<ChannelPanelWidget> {
...
+                    Row(children: _commandButtonsBuilder(commandsModel)),
```
kevmo314, line 289: I don't think the outer `Row` is necessary.

lib/models/commands.dart
```diff
@@ -0,0 +1,34 @@
...
+    _commands = [];
+  }
+
+  void addCommand(String command) {
+    if (_commands.contains(command)) {
+      _commands.remove(command);
+    }
+    if (_commands.length >= _maxNumberOfCommands) {
+      _commands.removeAt(
+          _maxNumberOfCommands - 2); // remove 2nd last, last is clear button
```
kevmo314, line 20: This is a bit odd, why is the clear button in the list of commands?

lib/models/commands.dart
```diff
@@ -0,0 +1,34 @@
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+
+class CommandsModel extends ChangeNotifier {
+  List<String> _commands = [];
```
kevmo314, line 5: `final`

lib/models/commands.dart
```diff
@@ -0,0 +1,34 @@
...
+      _commands.removeAt(
+          _maxNumberOfCommands - 2); // remove 2nd last, last is clear button
+    }
+    _commands.insert(0, command);
+    notifyListeners();
+  }
+
+  CommandsModel.fromJson(Map<String, dynamic> json) {
+    if (json['commands'] != null) {
+      final commands = List<String>.from(json['commands']);
```
kevmo314, line 28: Assign it directly

`_commands = List<String>...`

lib/models/commands.dart
```diff
@@ -0,0 +1,34 @@
...
+import 'package:flutter/material.dart';
+
+class CommandsModel extends ChangeNotifier {
+  List<String> _commands = [];
+  static const _maxNumberOfCommands = 10;
+
+  List<String> get commands => _commands;
+
+  void clear() {
+    _commands = [];
```
kevmo314, line 11:
```
_commands.clear();
notifyListeners();
```

kevmo314 (commented):
Code looks good, some changes to improve readability :)

lib/components/command_bar.dart
```diff
@@ -0,0 +1,61 @@
+import 'package:flutter/material.dart';
+import 'package:rtchat/models/commands.dart';
+import 'package:rtchat/models/channels.dart';
+import 'package:rtchat/models/adapters/actions.dart';
+
+class CommandBarWidget extends StatelessWidget {
+  // final void Function() onPressed;
+  final CommandsModel commandsModel;
```
kevmo314, line 8: Why not read the models in this widget?

lib/components/command_bar.dart
```diff
@@ -0,0 +1,61 @@
...
+  final ChannelsModel channelsModel;
+  final FocusNode chatInputFocusNode;
```
kevmo314, line 10: It's slightly more preferable to pass a callback instead of the entire focus node. This typically makes things easier to test.

```
final void Function() onUnfocus;
```

Then in the caller `CommandBarWidget(onUnfocus: () => chatInputFocusNode.unfocus());`

lib/components/channel_panel.dart
```diff
@@ -254,6 +262,18 @@ class _ChannelPanelWidgetState extends State<ChannelPanelWidget> {
           })
       : Container();
 
+  Widget _buildCommandBar(BuildContext context) {
+    final commandsModel = Provider.of<CommandsModel>(context, listen: false);
```
kevmo314, line 266: Use Consumers instead of Providers in the `build` functions:

```
Consumer<CommandModel>(builder: (context, commandModel, child) {
   ...
})
```

lib/components/command_bar.dart
```diff
@@ -0,0 +1,61 @@
...
+  }
+
+  List<TextButton> _commandButtonsBuilder(CommandsModel commandsModel) {
+    List<TextButton> commandButtons = [];
+    for (String command in commandsModel.commands) {
+      commandButtons.add(TextButton(
+        child: Text(command),
+        onPressed: () {
+          ActionsAdapter.instance
+              .send(channelsModel.subscribedChannels.first, command);
```
kevmo314, line 41: Since `channelsModel` is only used in a callback, this should be fetched via `Provider.of<...>` inside the callback.

lib/components/channel_panel.dart
```diff
@@ -254,6 +262,16 @@ class _ChannelPanelWidgetState extends State<ChannelPanelWidget> {
           })
       : Container();
 
+  Widget _buildCommandBar(BuildContext context) {
```
kevmo314, line 265: I would inline this function tbh

lib/components/command_bar.dart
```diff
@@ -0,0 +1,46 @@
...
+class CommandBarWidget extends StatelessWidget {
+  final void Function() chatInputUnfocus;
+  static const _commandBarHeight = 55.0;
+
+  const CommandBarWidget({Key? key, required this.chatInputUnfocus})
+      : super(key: key);
+
+  @override
+  Widget build(BuildContext context) {
+    return SizedBox(
```
kevmo314, line 16: Does this not autosize? ie the `SizedBox` shouldn't be necessary because the button bar has intrinsic size?

lib/components/command_bar.dart
```diff
@@ -0,0 +1,46 @@
...
+      height: _commandBarHeight,
+      child: Scrollbar(
+        isAlwaysShown: false,
```
kevmo314, line 19: Since the default of this param is false, what's the motivation to have a `Scrollbar` component?

lib/components/command_bar.dart
```diff
@@ -0,0 +1,46 @@
...
+        child: ListView(
```
kevmo314, line 20: I can see why you chose to use `Provider` down below, but that actually results in a bug: if the model changes without keyboard reopening, the change won't be reflected here. Instead, wrap the ListView in a Consumer and construct the items directly:

```
child: Consumer<CommandsModel>(builder: (context, commandsModel, child) {
  return ListView.builder(
    itemCount: commandsModel.commands.length,
    itemBuilder: (context, index) => TextButton(
      child: Text(commandsModel.commands[i]),
      onPressed: () { ... },
    ),
  );
}),

kevmo314 (approved):

kevmo314 (approved):

PR #201, opened 2021-08-27
feat: add prediction event to chat history

### LINKED ISSUE #155: Render predictions

juaoose:
Closes #155


lib/components/chat_history/twitch/prediction_event.dart
```diff
@@ -0,0 +1,109 @@
...
+      child: Stack(alignment: AlignmentDirectional.center, children: [
+        SizedBox(
+          height: 37,
+          child: LinearProgressIndicator(
+            value: outcome.points / max(1, totalPoints),
+            valueColor: AlwaysStoppedAnimation<Color>(outcome.widgetColor),
+            backgroundColor: outcome.widgetColor.shade200,
+          ),
+        ),
+        ...(isWinner
```
kevmo314, line 63: This might be a little more readable to use `Builder` but up to you.

```
Builder(builder: (context) {
  if (isWinner) {
    return Stack(children: [
     ...
    ]);
  } else {
    return Stack(children: [
      ...
    ]);
  }
});

kevmo314 (approved):

lib/components/chat_history/twitch/prediction_event.dart
juaoose: Yup, this conditional spread is not nice.

PR #217, opened 2021-08-30
feat: completed poll event config

### LINKED ISSUE #211: Polls settings page

chungwwei:
Closes #211


lib/models/messages/twitch/event.dart
```diff
@@ -100,7 +106,9 @@ class TwitchPollEventModel extends MessageModel {
         pollTitle: data['event']['title'],
         isCompleted: true,
         messageId: "poll${data['event']['id']}",
-        timestamp: data['timestamp'].toDate());
+        timestamp: data['timestamp'].toDate(),
+        startTimestamp: data['event']['started _at'].toDate(),
```
juaoose, line 110: `started _at` there is an extra space.

functions/package.json
```diff
@@ -31,4 +31,4 @@
     "typescript": "^4.3.5"
   },
   "private": true
-}
+}
```
juaoose, line 34: No real change here I guess?

lib/models/messages/twitch/event.dart
```diff
@@ -100,7 +106,9 @@ class TwitchPollEventModel extends MessageModel {
...
+        endTimestamp: data['event']['ended_at'].toDate());
```
juaoose, line 111: I guess the only thing that we might want to test with some data is if pin behaviour should change depending on the `status` property, maybe we don't want to pin in on of the two other status (archived and/or terminated)

juaoose (changes requested):

lib/models/messages/twitch/event.dart
chungwwei: I will make it auto unpin when the poll status is `archived` or `terminated`.

juaoose (approved):

PR #222, opened 2021-09-01
feat: completed channel point redemption event config

ryansng:
image
Appbar title `Channel Point Redemption Configuration` is too long to fit in the title therefore I used `Channel Point Configuration` instead.


ryansng:
If you are wondering why are some poll event stuffs being changed in the files it is because I tried to resolve the conflict on web and messed up so I pulled and resolved it locally.

lib/models/messages/twitch/eventsub_configuration.dart
```diff
@@ -175,6 +192,20 @@ class EventSubConfigurationModel extends ChangeNotifier {
...
+    notifyListeners();
+  }
+
+  setChannelPointRedemptionEventShowable(bool value) {
+    channelPointRedemptionEventConfig.showEvent = value;
+    notifyListeners();
+  }
+
+  setChannelPointRedemptionEventPinnnable(bool value) {
+    channelPointRedemptionEventConfig.isEventPinnable = value;
```
juaoose, line 206: Missing the object change notification `notifyListeners();`

juaoose (changes requested):
- Missing channel points in `toJson()`, which would prevent storing the settings on shared preferences.
- Also maybe we should think about how we want the pin behaviour to work when redemptions are cancelled or fulfilled (example: maybe if the redemption is cancelled we dont want it to pin)

ryansng:
> * Also maybe we should think about how we want the pin behaviour to work when redemptions are cancelled or fulfilled (example: maybe if the redemption is cancelled we dont want it to pin)

Currently there's 4 statuses `fulfilled`, `cancelled`, `unfulfilled` and `unknown`. Should I check when it is `cancelled` or `unfulfilled` and unpin it if it is still pinned?

juaoose:
> > * Also maybe we should think about how we want the pin behaviour to work when redemptions are cancelled or fulfilled (example: maybe if the redemption is cancelled we dont want it to pin)
> 
> Currently there's 4 statuses `fulfilled`, `cancelled`, `unfulfilled` and `unknown`. Should I check when it is `cancelled` or `fulfilled` and unpin it if it is still pinned?

I think something like that would work yes. (tbh I don't have a clear idea on how pins should work for these, because they vary from how each streamer handles reward fulfillment)

The only thing is that sometimes IRL streamers never fulfill some of the rewards so maybe it would be better to unpin unfulfilled rewards after some time?

What do you guys think? @chungwwei @kevmo314

kevmo314:
> > > * Also maybe we should think about how we want the pin behaviour to work when redemptions are cancelled or fulfilled (example: maybe if the redemption is cancelled we dont want it to pin)
> > 
> > 
> > Currently there's 4 statuses `fulfilled`, `cancelled`, `unfulfilled` and `unknown`. Should I check when it is `cancelled` or `fulfilled` and unpin it if it is still pinned?
> 
> I think something like that would work yes. (tbh I don't have a clear idea on how pins should work for these, because they vary from how each streamer handles reward fulfillment)
> 
> The only thing is that sometimes IRL streamers never fulfill some of the rewards so maybe it would be better to unpin unfulfilled rewards after some time?
> 
> What do you guys think? @chungwwei @kevmo314

I think we can reasonably assume there's two types of streamers: streamers who want all the rewards to auto-unpin and streamers who want to manually go through fulfillment.

The case of "some rewards require manual fulfillment but not others" seems to get quite complex so I would be ok with making that assumption.

ryansng:
image
The idea here is:
1. If status is `canceled`, event will not pin even if `pin duration > 0`
2. If status is `fulfilled`, event will pin for `pin duration` length
3. If status is `unfulfilled` and
    - `manualClear` is `false`, event will pin for `pin duration` length
    - `manualClear` is `true`, event will pin for 365 days 
4. If status changes from 'unfulfilled' to
   - 'fulfilled', event will pin for 'pin duration' length
   - 'canceled', event will not be pin for any longer (dismissed)

I have no way of confirming if any of these work as I do not know how to test 😅

juaoose:
> image
> The idea here is:
> 
> 1. If status is `canceled`, event will not pin even if `eventDuration > 0`
> 2. If status is `fulfilled`, event will pin for `eventDuration` length
> 3. If status is `unfulfilled` and
>    
>    * `manualClear` is `false`, event will pin for `eventDuration` length
>    * `manualClear` is `true`, event will pin for 365 days
> 4. If status changes from `unfulfilled` to
>    
>    * `fulfilled`, event will pin for `eventDuration` length
>    * `canceled`, event will not be pin for any longer (dismissed)
> 
> I have no way of confirming if any of these work as I do not know how to test 😅
> 
> Also, say if the broadcaster receives a redemption that requires fulfilment, the status is `unfulfilled` when he initially receives it, how does the broadcaster "fulfil" the redemption? Do I need to code a "swipe-to-fulfil" feature for the event?

Yeah the app doesn't curently allow you to fulfill them (afaik that can not be done via IRC, only through the API) and that might be nice eventually.

lib/components/chat_history/message.dart
```diff
@@ -1,4 +1,4 @@
-import 'package:firebase_remote_config/firebase_remote_config.dart';
+// import 'package:firebase_remote_config/firebase_remote_config.dart';
```
juaoose, line 1: Remote config code can be removed from here since it is no longer used for these events.

ryansng:
> > image
> > The idea here is:
> > 
> > 1. If status is `canceled`, event will not pin even if `eventDuration > 0`
> > 2. If status is `fulfilled`, event will pin for `eventDuration` length
> > 3. If status is `unfulfilled` and
> >    
> >    * `manualClear` is `false`, event will pin for `eventDuration` length
> >    * `manualClear` is `true`, event will pin for 365 days
> > 4. If status changes from `unfulfilled` to
> >    
> >    * `fulfilled`, event will pin for `eventDuration` length
> >    * `canceled`, event will not be pin for any longer (dismissed)
> > 
> > I have no way of confirming if any of these work as I do not know how to test 😅
> > Also, say if the broadcaster receives a redemption that requires fulfilment, the status is `unfulfilled` when he initially receives it, how does the broadcaster "fulfil" the redemption? Do I need to code a "swipe-to-fulfil" feature for the event?
> 
> Yeah the app doesn't curently allow you to fulfill them (afaik that can not be done via IRC, only through the API) and that might be nice eventually.
> 
> Also, my opinion is that maybe a shorter default for unfulfilled ones could be better, taking into account that we can't fulfill them from the app just yet.


I added a slider for additional pin duration, is this what you are looking for?
image

lib/components/chat_panel.dart
```diff
@@ -106,6 +107,21 @@ DateTime? _getExpiration(
       return null;
     }
     return model.endTimestamp.add(pollEventConfig.eventDuration);
+  } else if (model is TwitchChannelPointRedemptionEventModel) {
+    final channelPointRedemptionEventConfig =
+        eventSubConfigurationModel.channelPointRedemptionEventConfig;
+    final unfulfilledDuration =
+        channelPointRedemptionEventConfig.eventDuration +
+            channelPointRedemptionEventConfig.unfulfilledAdditionalDuration;
+    if (model.status == 'unfulfilled' && unfulfilledDuration > Duration.zero) {
```
kevmo314, line 116: Can we parse `status` into an enum in the model so we don't have to do string comparisons?

kevmo314 (approved):

kevmo314 (approved):

PR #223, opened 2021-09-02
feat: added hypetrain event config

chungwwei:


kevmo314 (commented):
Change looks good to me, but I'll let @juaoose officially review :)

juaoose (commented):
Nice, I believe the change to the model needs to be reflected in the widget test.

lib/models/messages/twitch/hype_train_event.dart
```diff
@@ -27,7 +31,9 @@ class TwitchHypeTrainEventModel extends MessageModel {
         level: data['event']['level'] ?? 1,
         progress: data['event']['progress'],
         goal: data['event']['goal'],
-        total: data['event']['total']);
+        total: data['event']['total'],
+        startTimestamp: data['event']['started_at'].toDate(),
```
juaoose, line 35: These have to be parsed since they're Strings.

juaoose (changes requested):

juaoose (approved):

PR #228, opened 2021-09-03
fix: removed redundant pin switch for events

### LINKED ISSUE #226: [events] remove isEventPinnable
If `eventDuration` is zero, it's semantically the same as not being pinnable, right? So I think we can get rid of `isEventPinnable` and make the sliders go to zero?

chungwwei:
Closes #226

Pin switch is redundant, its functionality can be replaced by the slider.

I also changed the divisions on the sliders for events, so they all range [0, 30] with a step of 2.


lib/components/chat_panel.dart
```diff
@@ -78,24 +78,24 @@ DateTime? _getExpiration(
     MessageModel model, EventSubConfigurationModel eventSubConfigurationModel) {
   if (model is TwitchRaidEventModel) {
     final raidEventConfig = eventSubConfigurationModel.raidEventConfig;
-    return raidEventConfig.isEventPinnable
+    return raidEventConfig.eventDuration.inSeconds > 0
```
kevmo314, line 81: `.eventDuration > Duration.zero` :)

juaoose (approved):

PR #245, opened 2021-09-08
feat: add settings page for command suggestion

ryansng:
- `Enable suggestion` default is `true`
- `Always on` default is `false`
- If `Enable suggestion` is `false`, `Always on` will be disabled

Edit: I renamed all "command bar" and "command suggestion" variables and texts to "command shortcut"

image


ryansng:
Added note `"!" commands are added automatically`
image

kevmo314 (commented):
Hmm I'm actually not so sure we should add a settings page for this. I think it's unobtrusive enough that it's fine to leave in always, especially since if you don't use commands like me then it won't show up anyways.

Perhaps one improvement we could do is adding a timestamp though so if you haven't used a command in, say, a week or so, it disappears?

ryansng:
Ok I will close this PR now

PR #246, opened 2021-09-08
style: event settings padding changes

ryansng:
Change the padding slightly to hopefully make it easier to see where each ListTile belong. Thoughts?

Another example with Divider in comment below

image


ryansng:
image

kevmo314 (approved):

PR #247, opened 2021-09-08
feat: kill commands after specified time passed

ryansng:
This only happens when app restarts


ryansng:
Tested this with `1 minute` time to live for each command and it works

juaoose (approved):

PR #248, opened 2021-09-08
fix: update channel points redemption event config padding

ryansng:


ryansng:
channel points redemption config was added after (https://github.com/muxable/rtchat/pull/246) so this should update its padding

chungwwei (approved):

PR #250, opened 2021-09-11
feat: option to blacklist a user in tts

chungwwei:
ig we can probably have a settings page for viewing the blacklist. A form for adding a name to the blacklist and swiping list item to remove from blacklist

image

image

### DIFF
--- a/lib/components/chat_history/message.dart
+++ b/lib/components/chat_history/message.dart
@@ -26,6 +26,7 @@ import 'package:rtchat/models/messages/twitch/prediction_event.dart';
 import 'package:rtchat/models/messages/twitch/subscription_event.dart';
 import 'package:rtchat/models/messages/twitch/subscription_gift_event.dart';
 import 'package:rtchat/models/messages/twitch/subscription_message_event.dart';
+import 'package:rtchat/models/tts.dart';
 import 'package:rtchat/models/user.dart';
 
 class ChatHistoryMessage extends StatelessWidget {
@@ -52,7 +53,7 @@ class ChatHistoryMessage extends StatelessWidget {
         if (loginChannel != viewingChannel) {
           return child;
         }
-
+        final ttsModel = Provider.of<TtsModel>(context, listen: false);
         return InkWell(
             onLongPress: () async {
               var showTimeoutDialog = await showDialog<bool>(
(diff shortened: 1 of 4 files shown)


lib/models/tts.dart
```diff
@@ -73,6 +74,20 @@ class TtsModel extends ChangeNotifier {
     notifyListeners();
   }
 
+  void addNameToBlacklist(String name) {
+    blacklist.add(name);
+    ttsHandler.blackList = blacklist;
```
kevmo314, line 79: I don't think this assignment is necessary

kevmo314 (commented):
Nice :)

lib/models/tts.dart
chungwwei: `ttsHandler` will not have the updated set if I don't use assignment to update the set.

lib/components/chat_history/message.dart
```diff
@@ -52,14 +53,34 @@ class ChatHistoryMessage extends StatelessWidget {
...
-
+        final ttsModel = Provider.of<TtsModel>(context, listen: false);
         return InkWell(
             onLongPress: () async {
               var showTimeoutDialog = await showDialog<bool>(
                   context: context,
                   builder: (context) {
                     return Dialog(
                       child: ListView(shrinkWrap: true, children: [
+                        if (!ttsModel.mutedUsers.contains(m.author)) ...[
```
kevmo314, line 64: Use `Builder` instead: https://api.flutter.dev/flutter/widgets/Builder-class.html

```
children: [
  Builder(builder: (context) {
    final ttsModel = Provider.of<...>(...);
    if (ttsModel.mutedUsers.contains(m.author)) {
      return ...;
    }
    return ...;
  }),
  ...
]
```

lib/models/tts.dart
```diff
@@ -73,6 +75,20 @@ class TtsModel extends ChangeNotifier {
     notifyListeners();
   }
 
+  void addNameToBlacklist(TwitchUserModel model) {
```
kevmo314, line 78: Why not just `mute`? `ttsModel.mute(m.author)` is pretty descriptive.

lib/models/tts.dart
kevmo314: I see. I think we could make this a passthrough call then, otherwise we're maintaining the muted users set in two places. ie do

```
ttsHandler.mutedUsers.add(user);
notifyListeners();
```

and don't maintain the set of muted users in the model. It's not actually strictly necessary.

lib/models/tts.dart
```diff
@@ -73,6 +74,19 @@ class TtsModel extends ChangeNotifier {
     notifyListeners();
   }
 
+  void mute(TwitchUserModel model) {
+    ttsHandler.mutedUsers.add(model);
+    notifyListeners();
+  }
+
+  void unmute(TwitchUserModel model) {
+    if (ttsHandler.mutedUsers.isNotEmpty &&
```
kevmo314, line 83: `Set.remove` returns whether or not the element was removed: https://api.dart.dev/stable/2.10.4/dart-core/Set/remove.html

So this can be:

```
if (ttsHandler.mutedUsers.remove(model)) {
  notifyListeners();
}
```

Actually the same is true for `add` :) https://api.dart.dev/stable/2.10.4/dart-core/Set/add.html

kevmo314 (approved):
Nice looks good!

PR #252, opened 2021-09-13
fix: don't use android queue handler for tts

kevmo314:


lib/models/messages/tts_audio_handler.dart
```diff
@@ -149,95 +132,103 @@ class TtsAudioHandler extends BaseAudioHandler with QueueHandler {
...
-  }
-
-  Future<void> skipToEnd() async {
-    await stop();
-    await skipToQueueItem(queue.value.length - 1);
   }
 
   Future<void> force(String message) async {
-    await skipToEnd();
+    index = messages.length;
```
juaoose, line 208: Should this be `messages.length -1`?

kevmo314: Whoops yeah good catch

juaoose (approved):

PR #263, opened 2021-09-22
feat: pin mod/vip messages

chungwwei:
image

### DIFF
--- a/lib/models/messages/twitch/message.dart
+++ b/lib/models/messages/twitch/message.dart
@@ -182,6 +182,10 @@ class TwitchMessageModel extends MessageModel {
 
   bool get isCommand => !isAction && message.startsWith("!");
 
+  bool get isVip => badges.any((element) => element.key == 'vip');
+
+  bool get isMod => badges.any((element) => element.key == 'moderator');
+
   List<MessageToken> get tokenized => _tokenized ??= tokenize();
   List<MessageToken>? _tokenized;
   List<MessageToken> tokenize() {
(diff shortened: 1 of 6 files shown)


lib/models/messages/twitch/message.dart
```diff
@@ -182,6 +182,10 @@ class TwitchMessageModel extends MessageModel {
 
   bool get isCommand => !isAction && message.startsWith("!");
 
+  bool get isVip => badges.any((element) => element.key == 'vip');
+
+  bool get isMod => badges.any((element) => element.key == 'moderator');
```
juaoose, line 187: I think IRC already sends a property telling if the user has the [mod badge](https://dev.twitch.tv/docs/irc/tags#usernotice-twitch-tags).

lib/screens/settings/chat_history.dart
```diff
@@ -164,6 +165,50 @@ class ChatHistoryScreen extends StatelessWidget {
                 }
               },
             ),
+            Consumer<TwitchMessageConfig>(builder: (context, model, child) {
+              return Padding(
+                padding: const EdgeInsets.all(16),
+                child: Column(
+                  crossAxisAlignment: CrossAxisAlignment.start,
+                  children: [
+                    Text("Pin mod message duration",
```
juaoose, line 174: More a question than a change suggestion: could it be better to use moderator instead of mod for more clarity? Could that be useful for less familiarized users?

juaoose (approved):

PR #266, opened 2021-11-28
fix: added cumulative months info in gifted sub

chungwwei:


kevmo314 (changes requested):
Neat thanks! `functions/package.json` seems kind of important though? 😅

PR #270, opened 2022-01-23
fix: wait for quicklink to be processed before adding more

juaoose:
Since adding a link involves fetching the metadata, the current validation was not effective considering that we're calling an async function.

I tried debouncing and this but this feels better... maybe when the quicklink is a slow website it might not be so nice.

Any ideas? Maybe using a different data structure could also work?


lib/screens/settings/quick_links.dart
```diff
@@ -121,13 +128,15 @@ class _QuickLinksScreenState extends State<QuickLinksScreen> {
...
       final url = _textEditingController.text;
+      _isAddingLink.value = true;
       final metadata = await MetadataFetch.extract(url);
 
       Provider.of<QuickLinksModel>(context, listen: false).addSource(
           QuickLinkSource(metadata?.title, _activeIcon, Uri.parse(url)));
 
       _textEditingController.clear();
       FocusScope.of(context).unfocus();
+      _isAddingLink.value = true;
```
kevmo314, line 139: Is this one supposed to be false?

lib/screens/settings/quick_links.dart
```diff
@@ -109,7 +110,13 @@ class _QuickLinksScreenState extends State<QuickLinksScreen> {
                     textInputAction: TextInputAction.done,
                     onEditingComplete: addLink),
               ),
-              IconButton(icon: const Icon(Icons.add), onPressed: addLink),
+              ValueListenableBuilder<bool>(
```
kevmo314, line 113: Hmm it might be worthwhile to extract this into a new component or call `setState()` here instead. Since the component is already stateful it's a bit odd to have another listener inside of it.

kevmo314 (commented):
Maybe we should take an alternate approach to the metadata, perhaps only store the links and fetch all the metadata every time the screen loads? In other words, the metadata is just to render the quicklinks on the settings page.

That would sort of solve the slow metadata issue I think.

lib/screens/settings/quick_links.dart
juaoose: I think setState requires a rerender of all the widget and ValueNotifier does not, I will look into extracting into a separate widget.

kevmo314 (approved):

juaoose:
Is it ok that the fetching occurs multiple times? Like whenever a new item is added, the whole fetch process happens again.

kevmo314:
That's probably okay. Those fetches are pretty cheap

kevmo314:
Does anything else need to be done for this PR?

PR #276, opened 2022-02-12
fix: trick fast drag webview

chungwwei:


kevmo314:
Does this include both the layout changes and the webview fix? Can we separate those into two separate PRs?

chungwwei:
The PR should have changes for the draggin only. Webview resizes on release.

lib/models/layout.dart
```diff
@@ -40,22 +40,39 @@ class LayoutModel extends ChangeNotifier {
...
   }
 
   double get panelHeight => _panelHeight;
 
   double get panelWidth => _panelWidth;
 
+  double get onDragStartHeight => _onDragStartHeight;
+
+  set onDragStartHeight(double onDragStartHeight) {
+    _onDragStartHeight = onDragStartHeight;
```
kevmo314, line 66: Is this missing a notifyListeners()?

lib/screens/home.dart
```diff
@@ -47,34 +47,63 @@ class _HomeScreenState extends State<HomeScreen> with TickerProviderStateMixin {
...
+        bool dragEnd = layoutModel.dragEnd;
+        Widget notifPanel = NotificationPanelWidget(
+          height: dragEnd
+              ? layoutModel.panelHeight.clamp(57, 500)
+              : layoutModel.onDragStartHeight.clamp(57, 500),
+        );
+        return Stack(
+          children: [
+            notifPanel,
+            // KeyboardVisibilityBuilder(builder: (context, isKeyboardVisible) {
```
kevmo314, line 65: Remove the commented section

lib/screens/home.dart
```diff
@@ -47,34 +47,63 @@ class _HomeScreenState extends State<HomeScreen> with TickerProviderStateMixin {
...
+                  height:
+                      _minimized ? 56 : layoutModel.panelHeight.clamp(56, 500),
+                ),
+                Expanded(
+                  child: Container(
+                    color: chathistoryBackground,
+                    child: DiscoWidget(
+                      isEnabled: widget.isDiscoModeEnabled,
+                      child: ChannelPanelWidget(
+                        onRequestExpand: () {
```
kevmo314, line 86: Can this hook be deleted?

kevmo314 (changes requested):

lib/screens/home.dart
chungwwei: `onRequestExpand` is the double tap functionality. The old one just brings the chat to the top, and you can't drag it until you double tap it again, bc `_minimized` is locking it to `panelHeight` 56.

1) I can put down `layoutModel.paneHeight = 56`, so that it brings chat to the top, and you can drag. The reason being I plan not to have the dragBar when the layout is locked in the next PR, so you can't double tap back.
2) keep the old one, and have the dragBar always, and you double tap to top and back when lcoked or unlocked.

which one should we do?

kevmo314: I think we should leave the drag bar for now, we can keep removing it in your next PR. Maybe useful would be to make the statistics bar widget the click target?

juaoose (approved):
I guess this behaviour would also be relevant for landscape mode, wouldn't it?

kevmo314 (approved):

PR #287, opened 2022-02-18
feat: new-ui look

chungwwei:
image

image


lib/components/channel_panel.dart
```diff
@@ -69,103 +59,27 @@ class _ChannelPanelWidgetState extends State<ChannelPanelWidget> {
...
+          if (layoutModel.locked) {
+            return Container();
+          }
+          return Center(
+            child: ClipRRect(
+              borderRadius: BorderRadius.circular(15.0),
+              child: Container(
+                width: 350,
+                height: 16,
+                color: Colors.blueAccent,
```
juaoose, line 72: I think we might want to use a color based on theme? How does the bar look on dark mode?

kevmo314: Yeah I think the Android one looks better. This is a bit tricky because it's not often we have this type of content where we need a split in apps so there might not be many good examples.

chungwwei:
image

Maybe having the viewer count widget on top of chat was a bad idea.
1) It covers pin events if there are any
2) makes dragging slower with the re-rendering

new proposal?:
1) nuke the quicklinks area, place the viewer count widget there.
2) quicklinks go to the side drawer if you swipe left on the screen?

juaoose:
image

KeN095:
> image

Would it be confusing to have two hamburger menu icons, one in webview and one in the app? Just realized it now after seeing the pics.

kevmo314:
> Would it be confusing to have two hamburger menu icons, one in webview and one in the app? Maybe a way to ease that is to add a subtle border around webview.

Yeah, let's switch the webview icon to something else. Since people use it primarily for notifications, maybe a bell? https://fonts.google.com/icons?selected=Material+Icons&icon.query=bell

chungwwei:
image
image

chungwwei:
drawer seems to be sluggish if the channel has alot of msgs. But, maybe the emulator is just slow?

kevmo314:
> drawer seems to be sluggish if the channel has alot of msgs. But, maybe the emulator is just slow?

Have you tried it on a real device? Otherwise this drawer looks like a good start!

lib/components/right_drawer.dart
```diff
@@ -0,0 +1,239 @@
...
+            child: _buildChannelListView(),
+          ),
+          const Divider(
+            indent: 30,
+            endIndent: 30,
+            thickness: 2,
+          ),
+
+          // quicklinks title
+          const Text("Quicklinks"),
```
kevmo314, line 65: I think we can omit the header actually.

lib/components/right_drawer.dart
```diff
@@ -0,0 +1,239 @@
...
+          // quicklinks
+          _buildQuicklinksListView(),
```
kevmo314, line 67: Create a new widget instead of builders.

lib/components/right_drawer.dart
```diff
@@ -0,0 +1,239 @@
...
+          BackdropFilter(
+            filter: ImageFilter.blur(sigmaX: 1, sigmaY: 2),
+            child: Container(
+              color: Colors.black.withOpacity(0),
+            ),
+          ),
+          const SizedBox(
+            height: 100,
+          ),
+          // channel
```
kevmo314, line 53: Use a DrawerHeader? https://api.flutter.dev/flutter/material/DrawerHeader-class.html

lib/components/right_drawer.dart
```diff
@@ -0,0 +1,239 @@
...
+            endIndent: 30,
+            thickness: 2,
+          ),
+
+          // setting
+          Consumer<LayoutModel>(builder: (context, layoutModel, child) {
+            if (layoutModel.locked) {
+              return ListTile(
+                leading: const Icon(Icons.lock_open_outlined),
+                title: const Text("lock Layout"),
```
kevmo314, line 80: Fix capitalization

lib/components/right_drawer.dart
```diff
@@ -0,0 +1,239 @@
...
+                onTap: () async {
+                  layoutModel.locked = !layoutModel.locked;
+                  Navigator.pop(context);
+                },
+              );
+            }
+            return ListTile(
+              leading: const Icon(Icons.lock_outline),
+              title: const Text("unlock Layout"),
```
kevmo314, line 89: Fix capitalization

lib/components/right_drawer.dart
```diff
@@ -0,0 +1,239 @@
...
+              onTap: () async {
+                layoutModel.locked = !layoutModel.locked;
+                Navigator.pop(context);
+              },
+            );
+          }),
+          ListTile(
+            leading: const Icon(Icons.refresh_outlined),
+            title: const Text("refresh audio sources"),
```
kevmo314, line 98: Fix capitalization

lib/components/right_drawer.dart
```diff
@@ -0,0 +1,239 @@
...
+    } else {
+      await launch(source.url.toString());
+    }
+  }
+
+  @override
+  Widget build(BuildContext context) {
+    print('rebuilding drawer');
+    return Drawer(
+      child: Column(
```
kevmo314, line 42: Use a ListView instead of a column, see https://docs.flutter.dev/cookbook/design/drawer#3-populate-the-drawer-with-items

kevmo314 (changes requested):

kevmo314:
> I tried `flutter run --profile` on a physical devise, better than `debug mode`, but still a little performance issue when I go to big streamer channel.
> 
> 1. weird that the setting page would lag when I enter a big streamer channel, idk the cause yet.
> 2. `ChatHistory` has slow performance when streamer is kinda big. All the messages are getting rebuilt/re-render, maybe we need to use `ListView` and [`ObjectKey`](https://api.flutter.dev/flutter/widgets/ObjectKey-class.html) to fix that in a separate issue.

Hmm does the performance issue happen without this change too? The list view should be reasonably optimized but maybe there's more room for improvement there.

lib/components/drawer/quicklinks_listview.dart
```diff
@@ -0,0 +1,52 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    return Consumer<QuickLinksModel>(
+        builder: (context, quickLinksModel, child) {
+      if (quickLinksModel.sources.isEmpty) {
+        return const Padding(
+          padding: EdgeInsets.all(16),
+          child: Center(
+            child: Text("Add quicklinks in setting"),
```
juaoose, line 31:
```suggestion
            child: Text("Add quicklinks in settings"),
```

lib/components/drawer/right_drawer.dart
```diff
@@ -0,0 +1,206 @@
...
+              final count = await audioModel.refreshAllSources();
+              ScaffoldMessenger.of(context).showSnackBar(SnackBar(
+                  content: Text(count == 1
+                      ? '1 audio source refreshed'
+                      : '$count audio sources refreshed')));
+            },
+          ),
+          ListTile(
+            leading: const Icon(Icons.settings_outlined),
+            title: const Text("Setting"),
```
juaoose, line 95:
```suggestion
            title: const Text("Settings"),
```

juaoose (changes requested):
There seems to be an horizontal overflow whenever viewer/follower count is "big".

lib/components/drawer/right_drawer.dart
```diff
@@ -0,0 +1,206 @@
...
+
+          const Divider(
+            indent: 30,
+            endIndent: 30,
+            thickness: 2,
+          ),
+
+          // setting
+          Consumer<LayoutModel>(builder: (context, layoutModel, child) {
+            if (layoutModel.locked) {
```
juaoose, line 61: I believe this one is flipped.
```suggestion
            if (!layoutModel.locked) {
```

lib/components/chat_panel.dart
```diff
@@ -4,6 +4,7 @@ import 'package:flutter/material.dart';
 import 'package:provider/provider.dart';
 import 'package:rtchat/components/chat_history/message.dart';
 import 'package:rtchat/components/pinnable/scroll_view.dart';
+import 'package:rtchat/components/statistics_bar.dart';
```
juaoose, line 7: Remove unused import.

juaoose (approved):

lib/components/channel_panel.dart
```diff
@@ -69,103 +59,33 @@ class _ChannelPanelWidgetState extends State<ChannelPanelWidget> {
...
-                onPressed: () {
-                  ttsModel.enabled = !ttsModel.enabled;
-                });
-          }),
-        ]),
+        child: Consumer<LayoutModel>(builder: (context, layoutModel, child) {
+          if (layoutModel.locked) {
+            return Container();
+          }
+          // return Center(
```
kevmo314, line 66: Remove commented code

kevmo314 (approved):

PR #304, opened 2022-03-16
feat: add host event to ui

### LINKED ISSUE #279: [events] add incoming hosts

juaoose:
Closes #279


juaoose:
I was not able to test the widget using NetworkImageWithRetry, seems like the timer would prevent the test from working... Ill try to read some more.

functions/src/index.ts
```diff
@@ -265,20 +265,24 @@ export const getStatistics = functions.https.onCall(async (data, context) => {
 export const getProfilePicture = functions.https.onRequest(async (req, res) => {
   const provider = req.query?.provider as string | null;
   const channelId = req.query?.channelId as string | null;
-  if (!provider || !channelId) {
+  const login = req.query?.login as string | null;
```
kevmo314, line 268: The host schema now includes channelId in the data, can you use that field instead?

juaoose: Isn't that the host channel info? The image is for the user that hosted

kevmo314: No channelId should be the hoster's channel. At least if it's not then that's a bug 😓

juaoose: thats confusing [considering](https://github.com/twitch-js/twitch-js/blob/476ff0ab0e5841836b49abb76bd27f90e3cb556a/src/Chat/__tests__/__snapshots__/chat.test.ts.snap#L1298)

kevmo314: Oh wait no you're right, the channelId is the hosted channel id. I think we should update the code though to include the hoster's channel id though: https://github.com/muxable/rtchat/blob/main/agent/src/agents/twitch.ts#L92

This would make the event consistent with the other ones

juaoose: So, call `https://api.twitch.tv/helix/users` when the agent receives the event to get the user id?

kevmo314: Yes

chungwwei (approved):

juaoose:
Updated the PR to no longer include the profile picture query by display name, since we plan on adding user ids for both the hoster (I propose `hosterUserId`) and the hosted channel `channelId`. @kevmo314

kevmo314:
@chungwwei can you take a look?

chungwwei (approved):

PR #307, opened 2022-03-19
fix: remove channelId

kevmo314:
Turns out it's pulling the wrong id right now this is the user's id oops.


juaoose:
Isnt channelId used to retrieve the messages?

kevmo314:
> Isnt channelId used to retrieve the messages?

Yeah but it's not the correct channel id. This is the firebase user id which doesn't match. We should actually call the twitch API here too.

juaoose (approved):

PR #308, opened 2022-03-19
feat: list viewers in a drawer

chungwwei:
image

### DIFF
--- a/lib/components/drawer/end_drawer.dart
+++ b/lib/components/drawer/end_drawer.dart
@@ -1,5 +1,4 @@
 import 'dart:convert';
-import 'dart:ui';
 
 import 'package:flutter/material.dart';
 import 'package:http/http.dart' as http;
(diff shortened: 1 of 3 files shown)


kevmo314:
Nice! Can we hide the header if there are no users in that section? Also post this to the discord when it gets in :)

KeN095:
Yup, it looks good! It might look better if icons appeared to the left of each header, but having just the header there is enough to portray the meaning.

KeN095 (approved):

juaoose (approved):
Should we remove the commented code?

lib/components/drawer/end_drawer.dart
```diff
@@ -0,0 +1,137 @@
...
+
+class LeftDrawerWidgetState extends State<LeftDrawerWidget> {
+  Future<dynamic>? viewersFuture;
+
+  _viewersFuture(String channel) async {
+    Uri uri = Uri.parse('https://tmi.twitch.tv/group/user/$channel/chatters');
+    final res = await http.get(uri);
+    if (res.statusCode == 200) {
+      List<String> broadcasterList =
+          List<String>.from(jsonDecode(res.body)['chatters']['broadcaster']);
```
kevmo314, line 25: Store `jsonDecode(req.body)` in a var so it only has to be decoded once

lib/components/drawer/end_drawer.dart
```diff
@@ -0,0 +1,137 @@
...
+  State<LeftDrawerWidget> createState() => LeftDrawerWidgetState();
+}
+
+class LeftDrawerWidgetState extends State<LeftDrawerWidget> {
+  Future<dynamic>? viewersFuture;
+
+  _viewersFuture(String channel) async {
+    Uri uri = Uri.parse('https://tmi.twitch.tv/group/user/$channel/chatters');
+    final res = await http.get(uri);
+    if (res.statusCode == 200) {
```
kevmo314, line 23: Use guard clauses instead, like put the error cases in the conditional. In other words,

```
if (res.statusCode != 200) {
  return {...}
}
...decode...
```

As a general rule, the last return should be the "success" case of the function.

PR #309, opened 2022-03-24
feat: searchbar to filter active viewers

chungwwei:
image

image


kevmo314:
🤯 Wow nice! Screenshot?

KeN095:
Looks great! Would be better to contain the search field within its own box give that box it's own background color.

lib/components/drawer/sliver_search_bar.dart
```diff
@@ -0,0 +1,56 @@
...
+import 'package:rtchat/models/drawer/endrawer/viewers_list_model.dart';
+
+class SliverSearchBarWidget extends StatefulWidget {
+  const SliverSearchBarWidget({Key? key}) : super(key: key);
+
+  @override
+  State<SliverSearchBarWidget> createState() => _SliverSearchBarWidgetState();
+}
+
+class _SliverSearchBarWidgetState extends State<SliverSearchBarWidget> {
```
kevmo314, line 12: This can be a StatelessWidget

lib/components/drawer/end_drawer.dart
```diff
@@ -66,66 +69,85 @@ class LeftDrawerWidgetState extends State<LeftDrawerWidget> {
...
-              slivers: [
-                // const SliverSearchBarWidget(),
-                SliverList(
-                  delegate: SliverChildListDelegate(
-                    const [
-                      SizedBox(
-                        height: 64,
-                      )
-                    ],
+            final viewersListModel =
```
kevmo314, line 72: Since this is the only place the viewer list is used in the app, it would be better to avoid using a model and do the filter in the widget.

lib/components/drawer/sliver_search_bar.dart
```diff
@@ -0,0 +1,51 @@
+import 'package:flutter/material.dart';
+
+class SliverSearchBarWidget extends StatelessWidget {
+  final Function filterBySearchBarText;
```
kevmo314, line 4: Type this as `Function(String)`.

Also prefix it with `on`, ie `onFilterBySearchBarText`. Event handlers tend to be prefixed with `on` in the standard lib.

kevmo314 (approved):
Minor comments, otherwise looks good :)

lib/components/drawer/end_drawer.dart
```diff
@@ -59,75 +128,74 @@ class LeftDrawerWidgetState extends State<LeftDrawerWidget> {
...
-            final moderatorList = data['moderators'];
-            final vipList = data['vips'];
-            final viewerList = data['viewers'];
+          if (snapshot.connectionState != ConnectionState.done) {
+            return const Center(child: CircularProgressIndicator());
+          } else {
             return CustomScrollView(
               slivers: [
-                // const SliverSearchBarWidget(),
+                searchBarWidget,
```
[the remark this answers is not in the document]
chungwwei: The value doesn't get erased, I had the widget as an instance variable so the searchBarWdiget doesn't rebuild when the user types into the textfield.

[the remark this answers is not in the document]
kevmo314: I see, that's actually okay to inline the widget then. Flutter is pretty good about efficient rebuilds, it's generally worth keeping the code more intuitive/cleaner than to optimize for rebuilds.

PR #318, opened 2022-04-04
fix: allow auth for account with no email

### LINKED ISSUE #317: can't login in the app

juaoose:
Fixes #317 

This is if we want to allow signup without email.


functions/src/auth.ts
```diff
@@ -108,7 +108,7 @@ app.get("/auth/twitch/callback", async (req, res) => {
 
   // save the profile information too.
   const twitchProfile = {
-    email,
+    ...email ? email: {},
```
kevmo314, line 111: Isn't email a string? Conditionally add the field?

kevmo314 (approved):

PR #331, opened 2022-04-16
fix: do not render activity feed preview if disabled

juaoose:
The app crashes whenever you open the activity feed while not signed in.


kevmo314 (approved):
Thanks!

PR #353, opened 2022-04-19
fix: set lowest default quality on stream preview

### LINKED ISSUE #264: [preview] stream preview should default to lowest quality
It looks like Twitch no longer supports the `&quality=` url param so this line no longer works. https://github.com/muxable/rtchat/blob/main/lib/components/notification_panel.dart#L101

To fix this, we'll need to instead switch to this embedded player: https://dev.twitch.tv/docs/embed/video-and-clips#interactive-frames-for-live-streams-and-vods and call `setQuality()` with the lowest quality parameter.

Probably the easiest way to do this would be to host a static html page and serve that via the [InAppWebView](https://inappwebview.dev/). I'm not sure how easy it is to host static pages though (looks like it can be done with [InAppLocalhostServer](https://inappwebview.dev/docs/in-app-localhost-server/basic-usage/)?), so it might be easier to dump it onto chat.rtirl.com/preview.html or something.

KeN095:
Sorry for the delay. Lowest default quality should be set to 160p. Height and width of the embed player still have hardcoded values and offline screen is still not implemented yet.

Closes #264


assets/twitch-player.html
```diff
@@ -0,0 +1,34 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <script src="https://player.twitch.tv/js/embed/v1.js"></script>
+    <title>Twitch Player</title>
+  </head>
+  <body>
+    <div id="twitch-embed" style="margin-left: -8px; margin-top: -8px"></div>
```
kevmo314, line 11: Can the width/height be determined by setting `style="width: 100%; height: 100%;"`?

assets/twitch-player.html
```diff
@@ -0,0 +1,34 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <script src="https://player.twitch.tv/js/embed/v1.js"></script>
+    <title>Twitch Player</title>
+  </head>
+  <body>
```
kevmo314, line 10: I think you can get rid of the `margin-left` and `margin-top` by removing the margin on the body and html tags: `style="margin: 0; padding: 0;"`

kevmo314 (commented):
Nice, just one question on the width/height, otherwise looks good :)

assets/twitch-player.html
KeN095: I believe I did that yesterday, the embed was looking a bit distorted and when I tried it again just now, it ended up looking like this: 
image

I want to achieve how it was done before, with player.twitch.tv occupying 100% of the whole page but I don't see any parameters from twitch docs that support that.

kevmo314: Ohh is it because the twitch player wants an explicit width/height in js? You can compute that by doing something like

```
      var width = document.getElementById('player').offsetWidth;
      var height = document.getElementById('player').offsetHeight;
```

KeN095: Width is fixed with that but not height: 
image

assets/twitch-player.html
KeN095: Done!

kevmo314 (approved):

PR #373, opened 2022-04-26
fix: set global font family

juaoose:
Only set a global font family instead of a complete theme.

There are a couple elements that are not using the theme so the font is not applied but I guess that can be resolved as part of the individual tasks we have for the new style (onboarding button and search box)

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -273,10 +273,12 @@ class _AppState extends State<App> {
       child: MaterialApp(
         title: 'RealtimeChat',
         theme: ThemeData(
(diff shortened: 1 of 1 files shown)


chungwwei (approved):

lib/main.dart
```diff
@@ -273,10 +273,12 @@ class _AppState extends State<App> {
       child: MaterialApp(
         title: 'RealtimeChat',
         theme: ThemeData(
+          fontFamily: GoogleFonts.poppins().fontFamily,
```
kevmo314, line 276: Nice I didn't know this could be set this way :D

PR #374, opened 2022-04-26
feat: better fit screen orientation toggle

juaoose:
This is probably personal opinion but the current size "feels" off.

Current:
Screenshot_20220426-184122

Proposed:
Screenshot_20220426-183748


kevmo314:
I like this change I think it looks better :)

KeN095 (approved):

PR #377, opened 2022-04-27
fix: add explicit sign in button and use bottom sheet

kevmo314:


juaoose (commented):
Should the new button have alignment for its text? Since its rather long it might not fit one line?

Somewhat related: Should we expand the bottom sheet to 100% height when keyboard is open? It looks weird on smaller devices.

I think this fixes the overflow I was experiencing on a smaller display.

juaoose:
> I think center aligned looks ok, what do you think?

Yeah center alignment I think is nice, I think right now it's to the left because mainAxisAlignment for Column is vertical, right?

kevmo314:
> Yeah center alignment I think is nice, I think right now it's to the left because mainAxisAlignment for Column is vertical, right?

I'm confused, can you point to the code change you're proposing? Isn't the column's alignment already set to center?

lib/screens/onboarding.dart
```diff
@@ -1,172 +1,76 @@
...
+                        padding: EdgeInsets.symmetric(horizontal: 64),
+                        child: SignInWithTwitch()),
+                  ),
+                  Text("or", style: Theme.of(context).textTheme.titleMedium),
+                  SizedBox(
+                    width: 400,
+                    child: Padding(
+                      padding: const EdgeInsets.symmetric(horizontal: 64),
+                      child: ElevatedButton(
+                        child: const Text("Continue without signing in"),
```
juaoose, line 49: Something like: or using Align
```suggestion
                        child: const Text(
                          "Continue without signing in",
                          textAlign: TextAlign.center,
                        ),
```

And I believe right now it's not horizontally centered because the property set to the parent Columns is dealing with [vertical alignment](https://docs.flutter.dev/development/ui/layout#aligning-widgets).

kevmo314: Ohhh I see ok, sounds good to me! Also fixed the alignment on tablet.

juaoose (approved):

PR #384, opened 2022-04-27
feat: prevent navigation in preview webview

juaoose:


kevmo314 (approved):
Oh wow that was simple, nice!

PR #386, opened 2022-05-01
fix: hide activity feed on disable

juaoose:
Currently, if you disable the activity feed while its open on the main page it is not hidden, so the only way to hide it is to use the stream preview button.


kevmo314 (approved):
Haha that's a funny bug.

PR #429, opened 2022-05-15
fix: improve unpinned messages rendering

### LINKED ISSUE #292: [perf] message widgets rebuild on new message
Verify if widgets for old messages are being rebuild every time we receive a new message.

kevmo314:
This is a pretty dramatic performance improvement. xqc's channel renders with very little lagging after this change. :)

Closes #292

### DIFF
--- a/lib/components/chat_panel.dart
+++ b/lib/components/chat_panel.dart
@@ -270,7 +270,12 @@ class _ChatPanelWidgetState extends State<ChatPanelWidget>
                           channel: widget.channel)),
                   isPinnedBuilder: (index) {
                     final expiration = expirations[index];
-                    return expiration != null && expiration.isAfter(now);
+                    if (expiration == null) {
+                      return PinState.notPinnable;
+                    }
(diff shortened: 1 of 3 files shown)


lib/components/chat_panel.dart
```diff
@@ -264,15 +264,18 @@ class _ChatPanelWidgetState extends State<ChatPanelWidget>
...
+                          message: messages[index],
+                          channel: widget.channel)),
                   isPinnedBuilder: (index) {
                     final expiration = expirations[index];
-                    if (expiration != null) {
-                      return expiration.isAfter(now);
+                    if (expiration == null) {
+                      return PinState.notPinnable;
                     }
-                    return false;
```
kevmo314, line 275: The gist of this improvement is that `null` used to represent a message that could not be pinned. Pinned messages are generally more expensive to layout and paint, but we were returning `false` instead of `null` which signaled that every message could have been pinned, thus dramatically increasing how expensive rendering is.

chungwwei (approved):

PR #439, opened 2022-05-21
feat: implement resilient network image

### LINKED ISSUE #403: create a more robust network image widget
We get an egregious number of crash logs due to failed network fetches from various images. We should create a new image widget ourselves that:

- Retries network requests automatically
- Caches images between invocations
- Includes a placeholder image when the image is not yet available

Turns out there isn't a dart widget already like this, perhaps if we make one we can publish it.

kevmo314:
Closes #403

### DIFF
--- a/lib/components/chat_history/twitch/follow_event.dart
+++ b/lib/components/chat_history/twitch/follow_event.dart
@@ -1,7 +1,6 @@
 // ignore_for_file: unused_import
 
 import 'package:flutter/material.dart';
-import 'package:flutter_image/network.dart';
 import 'package:rtchat/components/chat_history/decorated_event.dart';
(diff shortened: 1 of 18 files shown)


lib/components/chat_history/twitch/follow_event.dart
```diff
@@ -1,6 +1,8 @@
+// ignore_for_file: unused_import
```
juaoose, line 1: I think we dont need this one? Looks like all the imports are used.

juaoose (changes requested):

lib/components/chat_history/twitch/follow_event.dart
kevmo314: Good catch not sure where that came from

juaoose (approved):

PR #447, opened 2022-05-23
feat: label quicklink

### LINKED ISSUE #425: [Setting] ability to name URL quicklink

chungwwei:
Closes #425


lib/screens/settings/quick_links.dart
```diff
@@ -75,67 +69,87 @@ class _QuickLinksScreenState extends State<QuickLinksScreen> {
...
-                        return "This link already exists";
-                      }
-                      return null;
+              key: _formKey,
+              child: Column(children: [
+                TextFormField(
+                  controller: _labelEditingController,
+                  validator: (value) {
+                    if (value == null || value.isEmpty) {
+                      return 'Please enter a label';
```
kevmo314, line 78: I think we should default to the page's title if the label isn't specified instead of rejecting it. This lets the user add quicklinks as fast as possible.

lib/screens/settings/quick_links.dart
```diff
@@ -148,9 +156,14 @@ class _QuickLinksScreenState extends State<QuickLinksScreen> {
 
   void addLink() {
     if (_formKey.currentState!.validate()) {
+      if (_labelEditingController.text.isEmpty) {
+        _labelEditingController.text = "My Link";
```
kevmo314, line 160: No I mean we should fetch the metadata from the website like we did previously. For example if I enter google.com we can automatically figure out that it's titled Google.

lib/screens/settings/quick_links.dart
```diff
@@ -75,67 +69,81 @@ class _QuickLinksScreenState extends State<QuickLinksScreen> {
...
-                          (s) => s.url.toString() == Uri.encodeFull(value))) {
-                        return "This link already exists";
-                      }
-                      return null;
+              key: _formKey,
+              child: Column(children: [
+                TextFormField(
+                  controller: _labelEditingController,
+                  decoration: const InputDecoration(
+                    labelText: "Label",
```
kevmo314, line 77: I don't think we need to label it if we have the hint, it's a bit duplicitous:

image

lib/screens/settings/quick_links.dart
```diff
@@ -75,67 +69,81 @@ class _QuickLinksScreenState extends State<QuickLinksScreen> {
...
-                      final quickLinksModel =
-                          Provider.of<QuickLinksModel>(context, listen: false);
-                      if (quickLinksModel.sources.any(
-                          (s) => s.url.toString() == Uri.encodeFull(value))) {
-                        return "This link already exists";
-                      }
-                      return null;
+              key: _formKey,
+              child: Column(children: [
+                TextFormField(
```
kevmo314, line 74: This is kinda subtle but can we hide the label field if the url is empty? The user should enter the URL first anyways so this would keep it clean when they're not adding quick links.

lib/screens/settings/quick_links.dart
kevmo314: Actually, the hint should be the auto-fetched metadata name if possible. For example, if I enter google.com as the URL the label hint should be "Google"

kevmo314:
Updated the behavior of this PR to be a little more intuitive I think if anyone wants to give it a shot

juaoose (commented):
Does this PR require some migration code for existing quick links?

kevmo314:
> Does this PR require some migration code for existing quick links? @kevmo314

Good catch, added a migration path in the json parsing.

PR #462, opened 2022-05-27
fix: improve prediction event legibility

cswills:
Improves the legibility of prediction events in dark mode

Before:
image

After:
image


kevmo314 (approved):
Nice!

PR #469, opened 2022-05-31
feat: add endpoint for cloud tts

kevmo314:


functions/src/tts.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  const response = await fetch(
+    "https://texttospeech.googleapis.com/v1/text:synthesize",
+    {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json; charset=utf-8",
+        Authorization: `Bearer ${token}`,
+      },
+      body: JSON.stringify({
```
juaoose, line 39: I guess it defaults to english?

functions/src/tts.ts
```diff
@@ -0,0 +1,53 @@
+import * as functions from "firebase-functions";
+// import * as admin from "firebase-admin";
```
juaoose, line 2:
```suggestion
```

PR #505, opened 2022-06-16
feat: force interaction lock

### LINKED ISSUE #498: [layout] remove disable interaction option
Right now if the disable interaction option is disabled, locking the layout doesn't really do much. We should remove that option and make locking the layout disable interaction always.

kevmo314:
Closes #498.

Actually not 100% sure this is the right choice, what do you think?

### DIFF
--- a/lib/components/chat_history/message.dart
+++ b/lib/components/chat_history/message.dart
@@ -48,7 +48,7 @@ class ChatHistoryMessage extends StatelessWidget {
           padding: const EdgeInsets.symmetric(horizontal: 16),
           child: TwitchMessageWidget(m),
         );
-        if (layoutModel.isInteractionLockable && layoutModel.locked) {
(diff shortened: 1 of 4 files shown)


juaoose (commented):
Didn't you use this? Otherwise I don't know of many streamers having the phone directly under the rain...

kevmo314:
> Didn't you use this? Otherwise I don't know of many streamers having the phone directly under the rain...

Yeah I used this, but I think locking the layout really doesn't provide much value with the new UI if this isn't enabled. Maybe we should remove the lock layout button?

juaoose:
The only drawback is that the element right below the "resize bar" is a scrollable list, but its hard to tell if its really annoying since streamers maybe dont really scroll chat that much? I guess we could try whichever approach and if it really matters there will be feedback?

juaoose (approved):

PR #506, opened 2022-06-16
chore: wire link-preview-widget

chungwwei:
seems like re-renders still happen when you scroll up and down.

 when viewport changes, the preview card would re-render and would that be a concern?

image


kevmo314:
I think the rerendering on scroll is ok. We should, however, ellipsize the title and description so they don't span more than one line. Maybe the description can be two.

juaoose (approved):
Maybe we should change `cardColor` to a different color, as the current black makes link previews look plain in dark theme. Maybe `ThemeColors.detailColor` works?

PR #521, opened 2022-06-27
feat: render replies

### LINKED ISSUE #305: render replies
We get data on who a user is replying to, we should render some reply indicator

image

image

cswills:
closes #305 
image

Ellipsizing is weird in flutter as it cuts off whole words. It is mainly a problem when using large font sizes.
flutter/flutter#18761

### DIFF
--- a/lib/models/messages/twitch/message.dart
+++ b/lib/models/messages/twitch/message.dart
@@ -3,6 +3,7 @@ import 'package:linkify/linkify.dart';
 import 'package:rtchat/models/messages/message.dart';
 import 'package:rtchat/models/messages/tokens.dart';
 import 'package:rtchat/models/messages/twitch/emote.dart';
+import 'package:rtchat/models/messages/twitch/reply.dart';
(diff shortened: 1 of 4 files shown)


lib/models/messages/twitch/message.dart
```diff
@@ -161,6 +161,7 @@ Iterable<MessageToken> rootEmoteTokenizer(String message, String emotes) sync* {
 class TwitchMessageModel extends MessageModel {
   final TwitchUserModel author;
   final String message;
+  final Map<String, dynamic>? reply;
```
kevmo314, line 164: Can we parse `reply` into a typed object instead of leaving it as a map?

Something like

```
class TwitchMessageReplyModel {
  final String messageId;
  final String message;
  final TwitchUserModel author;
}
```

cswills: Done!

kevmo314 (approved):

PR #524, opened 2022-06-28
fix: change chin to background color

### LINKED ISSUE #523: [chat_history] chin on home screen is blue
The chin on iPhone 13 Max is blue, looks like it's still the accent color.

cswills:
closes #523?
I don't have an iPhone to test on, but this seems to be the cause of the issue.
https://github.com/muxable/rtchat/blob/7a03b5e8b3ec48aa6f0d2eaee2a9857fb7571adb/lib/screens/home.dart#L228-L232


KeN095:
Seems fixed to me. 
Simulator Screen Shot - iPhone 13 - 2022-06-28 at 19 13 14

KeN095 (approved):

PR #525, opened 2022-06-28
feat: use custom dark theme for onboarding screen

cswills:
Currently the onboarding screen uses the default Material dark theme which uses a slightly different blue primary and teal secondary color.


lib/themes.dart
```diff
@@ -0,0 +1,82 @@
...
+Color tintColor(Color color, double factor) => Color.fromRGBO(
+    tintValue(color.red, factor),
+    tintValue(color.green, factor),
+    tintValue(color.blue, factor),
+    1);
+
+final primarySwatch = generateMaterialColor(Themes.accentColor);
+
+class Themes {
+  static const accentColor = Color(0xFF009FDF);
```
kevmo314, line 33: I'm not so sure storing the color here is a good idea since it makes for two ways to access the same color. The ThemeData seems legit though.

juaoose (approved):

PR #533, opened 2022-06-29
fix: check profile exists before get

kevmo314:
https://console.firebase.google.com/u/0/project/rtchat-47692/crashlytics/app/android:com.rtirl.chat/issues/d8b958a8c356331fa06e2138539b7d5f?time=last-seven-days&versions=2.2.2%20(4518243)&sessionEventKey=62B9E40A01DB00012323C00FDE264A53_1692413470073385142

### DIFF
--- a/lib/models/adapters/profiles.dart
+++ b/lib/models/adapters/profiles.dart
@@ -13,11 +13,13 @@ class ProfilesAdapter {
   Stream<Channel?> getChannel(
       {required String userId, required String provider}) {
     return db.collection("profiles").doc(userId).snapshots().map((event) {
-      final data = event.get("twitch");
-      if (data != null) {
-        return Channel(provider, data['id'], data['displayName']);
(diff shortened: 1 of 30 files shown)


lib/models/adapters/profiles.dart
```diff
@@ -13,11 +13,13 @@ class ProfilesAdapter {
   Stream<Channel?> getChannel(
       {required String userId, required String provider}) {
     return db.collection("profiles").doc(userId).snapshots().map((event) {
-      final data = event.get("twitch");
-      if (data != null) {
-        return Channel(provider, data['id'], data['displayName']);
+      if (event.exists) {
```
juaoose, line 16: Don't we want `!event.exists`?

kevmo314: Haha good catch. Tired 💀

juaoose (approved):

PR #538, opened 2022-06-30
feat: chat-mode via dialog

chungwwei:
image

### DIFF
--- a/lib/components/message_input.dart
+++ b/lib/components/message_input.dart
@@ -4,6 +4,7 @@ import 'package:provider/provider.dart';
 import 'package:rtchat/components/emote_picker.dart';
 import 'package:rtchat/models/adapters/actions.dart';
 import 'package:rtchat/models/channels.dart';
+import 'package:rtchat/models/chat_mode.dart';
 import 'package:rtchat/models/commands.dart';
 
 class MessageInputWidget extends StatefulWidget {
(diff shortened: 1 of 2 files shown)


kevmo314:
Hmm, this is probably a good candidate for the slash command UI you mentioned that Discord has. I think having a dedicated button might be too intrusive, but we could trigger it if the user types a slash?

lib/components/message_input.dart
```diff
@@ -20,6 +23,89 @@ class _MessageInputWidgetState extends State<MessageInputWidget> {
...
+    if (lst.isEmpty) {
+      hideOverlay();
+      return;
+    }
+
+    final lstSize = lst.length;
+    const listTileSize = 75; // the roughly size of a listTile
+    final shiftUp = min(lstSize * listTileSize, 300);
+
+    entry = OverlayEntry(builder: (context) {
```
kevmo314, line 74: Can we make the state whether or not the widget is visible and render this in the `build()` function? Leaving the Widget construction outside `build()` can be a bit dangerous.

kevmo314 (changes requested):

kevmo314:
Just patched this in and tried it, it's pretty cool! I think we can probably add some ! commands as well in the future and integrate this with the ! command bar that @juaoose added somehow, but this is a great start.

lib/models/chat_mode.dart
```diff
@@ -0,0 +1,21 @@
+class ChatMode {
```
kevmo314, line 1: Make this an enum: https://dart.dev/guides/language/language-tour#enumerated-types

lib/components/message_input.dart
chungwwei: I'm not sure if I can move the `Overlay` widget into the `build()`. 
Rendering a `Overlay` seems to require insertion and removal for delete.

kevmo314 (approved):

PR #539, opened 2022-06-30
feat: enable edge-to-edge

cswills:
Sets the system navigation bar to be transparent


kevmo314 (approved):

kevmo314:
Does this still look reasonable with the layout locked?

cswills:
> Does this still look reasonable with the layout locked?

Yes. The messages don't scroll under the navbar.

PR #541, opened 2022-07-01
feat: increase number of divisions on font size

kevmo314:
Someone requested the ability to have 18px font.


juaoose (approved):
That's very specific haha

PR #543, opened 2022-07-01
feat: render /clear events

kevmo314:


lib/models/messages/message.dart
```diff
@@ -21,3 +21,7 @@ class SystemMessageModel extends MessageModel {
   SystemMessageModel({required this.text})
       : super(messageId: "", timestamp: DateTime.now());
 }
+
+class ChatClearedEventModel extends MessageModel {
+  ChatClearedEventModel() : super(messageId: "", timestamp: DateTime.now());
```
juaoose, line 26: Is it important for `messageId` to have a value since its used in a `Key()` or not really?

kevmo314: It's actually ok. The worst case scenario is that the subtree widget gets rebuilt.

juaoose (approved):

PR #550, opened 2022-07-08
feat: allow polls and predictions to be expanded

juaoose:


kevmo314:
Screenshot? :o

juaoose:
Screenshot_20220710-141050
Screenshot_20220710-141055

kevmo314:
Can we add a poll summary line too in the collapsed view? So the streamer can see the winning option without expanding. Doing it in a followup PR is fine too.

chungwwei (approved):

PR #552, opened 2022-07-11
feat: add connection status bar when disconnected

### LINKED ISSUE #38: [chat_history] online/offline status
While message delivery is pretty robust right now, it's hard to tell if messages are being delayed due to lack of connection. A message that shows up when you're offline would be nice.

kevmo314:
Closes #38

### DIFF
--- a/lib/components/connection_status.dart
+++ b/lib/components/connection_status.dart
@@ -2,12 +2,6 @@ import 'package:firebase_database/firebase_database.dart';
 import 'package:flutter/material.dart';
 import 'package:rtchat/components/chat_panel.dart';
 
-Stream<DateTime?> lastDisconnectTime() {
-  return FirebaseDatabase.instance.ref('.info/connected').onValue.map((event) {
-    return event.snapshot.value == false ? DateTime.now() : null;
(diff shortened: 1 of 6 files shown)


lib/components/connection_status.dart
```diff
@@ -0,0 +1,41 @@
+import 'package:firebase_database/firebase_database.dart';
+import 'package:flutter/material.dart';
+import 'package:rtchat/components/chat_panel.dart';
+
+Stream<DateTime?> lastDisconnectTime() {
+  return FirebaseDatabase.instance.ref('.info/connected').onValue.map((event) {
```
juaoose, line 6: Does the connection to Firebase DB gets reestablished? Since we don't have any client to that, or do we?

juaoose: Tried it out and the widget seems to leave for a bit and then come back forever :(

kevmo314: Ok I added a sentinel watcher, it's a bit hacky but seems to be more reliable...

juaoose (approved):

PR #562, opened 2022-07-16
fix: horizontal drag without opening drawer

### LINKED ISSUE #555: Command bar not scrollable
https://github.com/muxable/rtchat/blob/main/lib/screens/home.dart#L174
https://github.com/muxable/rtchat/blob/main/lib/components/message_input.dart#L132

Swiping opens the drawers, but it's also the same gesture to scroll through commands. 
Flutter seems to prioritize opening the drawers in the current codebase.

cswills:
closes #555 

This was a lot more difficult than I thought it would be. To allow for resizing of the activity feed in landscape mode, a custom gesture is used that has priority in the gesture arena. However, this method does not work for the command bar because the gesture has priority over the button presses as well. I found a workaround though. The command bar only shows when the keyboard is open, so disabling the drawer open gesture while the keyboard is open allows the user to scroll through commands.


kevmo314:
> closes #555
> 
> This was a lot more difficult than I thought it would be. To allow for resizing of the activity feed in landscape mode, a custom gesture is used that has priority in the gesture arena. However, this method does not work for the command bar because the gesture has priority over the button presses as well. I found a workaround though. The command bar only shows when the keyboard is open, so disabling the drawer open gesture while the keyboard is open allows the user to scroll through commands.

Hmm, would this work if we show the command bar/other sliding gestures without the keyboard? @chungwwei's upcoming UI might result in that. I suppose a way around that is we could avoid horizontal drag patterns in the UI.

cswills:
> 
> Hmm, would this work if we show the command bar/other sliding gestures without the keyboard? @chungwwei's upcoming UI might result in that. I suppose a way around that is we could avoid horizontal drag patterns in the UI.

This method only works with widgets whose only gesture is dragging, but I could look into this more for an alternative method for a persistent command bar. With the current solution, either the command bar would have to scroll vertically (maybe collapsible?), or the drawer gesture detector's width would have to be reduced. For the latter, extra horizontal padding could be added to the command bar for an easier to open drawer.

juaoose:
Maybe we should just revert to the original behavior for a while? While having this is nice, maybe it isn't that relevant? wdyt @cswills @kevmo314

kevmo314 (commented):
I thought about this some more, I think the horizontal drag to open the drawer is actually really nice. I'd rather relayout the command bar than lose that feature. So actually let's:

1. Skip this PR (sorry @cswills D:) and leave the bug as-is for now
2. In @chungwwei's upcoming feature we'll also incorporate the commands in a vertically scrolling grid or something, on a separate tab besides slash commands. I'll doodle some UI flows for this.

cswills:
> I thought about this some more, I think the horizontal drag to open the drawer is actually really nice. I'd rather relayout the command bar than lose that feature. So actually let's:
> 
> 1. Skip this PR (sorry @cswills D:) and leave the bug as-is for now
> 2. In @chungwwei's upcoming feature we'll also incorporate the commands in a vertically scrolling grid or something, on a separate tab besides slash commands. I'll doodle some UI flows for this.

Should I create another PR with just the fix for resizing the activity feed in landscape mode?

kevmo314:
> @kevmo314 Should I create another PR with just the fix for resizing the activity feed in landscape mode?

Yeah sounds good

PR #564, opened 2022-07-21
feat: language select screen

cswills:


kevmo314 (commented):
Can we use broadcaster_language? It would be cool if this didn't have to be configured, it could be wired through getUserStatistics https://dev.twitch.tv/docs/api/reference#get-channel-information

cswills:
I didn't realize tts could be enabled on other channels. Now the language selection screen is only visible if the broadcaster's language is set to asl or other.

kevmo314 (approved):

PR #570, opened 2022-07-27
feat: exclaimation mark commands overlay scroll

chungwwei:
image

image


kevmo314:
Maybe we can make it so the ! commands show by default and line them up like this?

image

That is, make the command bar wrap. That way we're not duplicating the command bar and we can get rid of it.

kevmo314:
I think we can reduce the padding around the chips and probably we don't need the background, but this is a good start. I'll send over some UI tweaks after this gets in.

kevmo314 (approved):

PR #599, opened 2022-08-10
feat: add cloud tts

### LINKED ISSUE #580: [tts] integrate cloud tts with current tts system

cswills:
closes #580

### DIFF
--- a/lib/models/tts.dart
+++ b/lib/models/tts.dart
@@ -1,20 +1,25 @@
 import 'dart:async';
 import 'dart:collection';
+import 'dart:convert';
 import 'dart:io';
 
 import 'package:cloud_functions/cloud_functions.dart';
+import 'package:crypto/crypto.dart';
 import 'package:firebase_crashlytics/firebase_crashlytics.dart';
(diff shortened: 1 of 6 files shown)


lib/models/tts.dart
```diff
@@ -283,13 +303,45 @@ class TtsModel extends ChangeNotifier {
...
+        final response =
+            await FirebaseFunctions.instance.httpsCallable("synthesize")({
+          "voice": voice ?? "en-US-WaveNet-F",
+          "language": _language.languageCode,
+          "text": vocalization,
+          "rate": _speed * 1.5 + 0.5,
+          "pitch": pitch ?? 0,
+        });
+        final bytes = const Base64Decoder().convert(response.data);
+        final audioPlayer = AudioPlayer();
```
juaoose, line 341: Does `AudioPlayer` make sense as a class member? Just like `FlutterTts`, that way we don't need to initialize on every `say()` call, also we could dispose on the `ChangeNotifier`s dispose.

juaoose (approved):

PR #606, opened 2022-08-16
feat: use custom overlay for stream preview

### LINKED ISSUE #545: improve web preview controls
The size of the webview makes it hard to interact with twitch's embedded player, it would be nice ifwe could build an overlay on top of the webview to improve player controls.

kevmo314:
Closes #545, #414


KeN095 (approved):

KeN095:
Since we're back to the regular twitch URL, shouldn't `twitch-player.html` be deleted?

PR #609, opened 2022-08-17
chore: clean up text input and emote picker

kevmo314:
This removes the overlay behavior but tbh I think it's better without the overlay.

### DIFF
--- a/lib/components/message_input.dart
+++ b/lib/components/message_input.dart
@@ -39,36 +39,35 @@ class _MessageInputWidgetState extends State<MessageInputWidget> {
   final _textEditingController = TextEditingController();
   final _chatInputFocusNode = FocusNode();
   var _isEmotePickerVisible = false;
-  late StreamSubscription<bool> keyboardSubscription;
+  var _isTextCommand = false;
+  var _isKeyboardVisible = false;
+  late StreamSubscription keyboardSubscription;
   var _emoteIndex = Random().nextInt(_emotes.length);
-  OverlayEntry? entry;
 
   @override
   void initState() {
     super.initState();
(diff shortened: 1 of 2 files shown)


lib/components/message_input.dart
```diff
@@ -206,30 +90,75 @@ class _MessageInputWidgetState extends State<MessageInputWidget> {
...
-          var initialText = _textEditingController.text;
-          if (initialText.isNotEmpty) {
-            _textEditingController.text =
-                initialText.substring(0, initialText.length - 1);
-          }
-        },
         onEmoteSelected: (emote) {
-          _textEditingController.text =
-              "${_textEditingController.text} ${emote.code}";
+          setState(() {
```
juaoose, line 100: Won't this and the listener cause rerenders when using the emote picker?

lib/components/message_input.dart
```diff
@@ -288,22 +217,16 @@ class _MessageInputWidgetState extends State<MessageInputWidget> {
...
                     if (filtered == text) {
                       return;
                     }
-                    _textEditingController.value = TextEditingValue(
-                        text: filtered,
-                        selection: TextSelection.fromPosition(TextPosition(
-                            offset: _textEditingController.text.length)));
+                    setState(() {
+                      _textEditingController.value = TextEditingValue(
+                          text: text,
```
juaoose, line 226:
```suggestion
                          text: filtered,
```

juaoose (changes requested):

lib/components/message_input.dart
kevmo314: Yes but we have to trigger a rerender to collapse the command panel

juaoose (approved):

PR #635, opened 2022-08-23
fix: disable overlay when video is not playing, add loading overlay

### LINKED ISSUE #375: offline screen for stream preview

kevmo314:
Closes #375

### DIFF
--- a/lib/components/stream_preview.dart
+++ b/lib/components/stream_preview.dart
@@ -67,7 +67,6 @@ class _StreamPreviewState extends State<StreamPreview> {
                 ?.runJavascript("action(Actions.SetQuality, '160p')");
           }
         },
-        backgroundColor: Colors.transparent,
         javascriptMode: JavascriptMode.unrestricted,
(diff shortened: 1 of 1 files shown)


lib/components/stream_preview.dart
```diff
@@ -76,17 +76,17 @@ class _StreamPreviewState extends State<StreamPreview> {
               name: "Flutter",
               onMessageReceived: (message) {
                 final params = jsonDecode(message.message)["params"];
-                if (params != null) {
+                if (params is Map) {
                   setState(() => _playerState = params["playback"]);
```
juaoose, line 80: Not exactly from this PR, but it looks like this `setState((){})` will throw errors as it's sometimes called then the widget is not mounted. It does not seem to crash the app.

juaoose (approved):

lib/components/stream_preview.dart
kevmo314: Oh good catch, I added some mounted guards to catch that.

PR #636, opened 2022-08-23
feat: improve activity feed settings and url resolution

kevmo314:


lib/screens/settings/activity_feed.dart
```diff
@@ -121,9 +125,16 @@ class _ActivityFeedScreenState extends State<ActivityFeedScreen> {
                                         });
                                   });
                             })),
+                    validator: (value) {
+                      if (value == null || Uri.tryParse(value) != null) {
```
juaoose, line 129:
```suggestion
                      if (value == null || Uri.tryParse(value) == null) {
```

juaoose (approved):

lib/screens/settings/activity_feed.dart
kevmo314: Actually the first null check was also wrong, thanks for catching that.

PR #638, opened 2022-08-23
feat: improve stream preview icons visibility

cswills:
In light mode, the stream preview buttons are black, blending in to the dark background


kevmo314 (approved):
Thanks!

PR #643, opened 2022-08-25
fix: improve rendering performance

kevmo314:


lib/components/pinnable/scroll_view.dart
```diff
@@ -68,6 +70,8 @@ class PinnableMessageScrollView extends ScrollView {
       }
       final pinned = isPinnedBuilder(nextPinnableIndex);
       final sliver = PinnableMessageSliver(
+        // key by first and last message id.
+        key: ValueKey("$start-$nextPinnableIndex"),
```
juaoose, line 74: wont these keys be the same?

kevmo314: Fixed

juaoose (approved):

PR #651, opened 2022-08-26
fix: use a more stable key

kevmo314:


lib/components/pinnable/scroll_view.dart
```diff
@@ -70,8 +71,7 @@ class PinnableMessageScrollView extends ScrollView {
       }
       final pinned = isPinnedBuilder(nextPinnableIndex);
       final sliver = PinnableMessageSliver(
-        // key by index.
-        key: ValueKey(start),
+        key: key,
```
juaoose, line 74: Doesn't using the same key conflict when there is a pinned element?

kevmo314: Oh good catch, I missed that the conditional didn't have a `continue`.

PR #664, opened 2022-08-29
feat: fn to store address

chungwwei:


functions/src/alchemy_webhook.ts
```diff
@@ -0,0 +1,44 @@
...
+  res.status(200).send();
+});
+
+
+export const setRealTimeCashAddress = functions.https.onCall(async (data, context) => {
+  if (!context.auth) {
+    throw new functions.https.HttpsError("permission-denied", "missing auth");
+  }
+	// console.log("data", data);
+	const userId = data?.userId;
```
kevmo314, line 23: `context.auth.uid`, we should never trust user input.

```suggestion
	const userId = context.auth.uid;
```

kevmo314 (changes requested):

functions/src/alchemy_webhook.ts
chungwwei: https://github.com/muxable/rtchat/blob/4e4bd92591f6d172dd9c187f8eca6bd0717cc1ce/lib/models/adapters/donations.dart#L29-L32

but the function in adapter does take in a `userId` from `UserModel`. Is `userId` the same as `context.auth.uid`?

kevmo314: Yes it is, actually that's a mistake, the `userId` doesn't need to be passed to that function.

https://github.com/muxable/rtchat/blob/617c57ea4b0a41aec8c10ca409cbea37e00c0bda/lib/screens/settings/donations.dart#L18

kevmo314 (approved):
Can you run `prettier` on the ts file?

PR #677, opened 2022-09-06
chore: migrate emotes to new function

kevmo314:


test/models/messages/twitch/message_test.dart
```diff
@@ -114,21 +114,28 @@ void main() {
     });
 
     test('third party emotes should tokenize', () {
-      final source = Uri.parse("https://3pemote");
+      final source = "https://3pemote";
```
juaoose, line 117:
```suggestion
      const source = "https://3pemote";
```

test/models/messages/twitch/message_test.dart
```diff
@@ -199,18 +206,25 @@ void main() {
     });
 
     test('third party emotes and emojis tokenize', () {
-      final source = Uri.parse("https://3pemote");
+      final source = "https://3pemote";
```
juaoose, line 209:
```suggestion
      const source = "https://3pemote";
```

juaoose (approved):

juaoose:
Looks like we either have to grab the small emote for some of the 3rd party ones or limit the size of the emotes when displayed.
Screenshot_20220906-133502

kevmo314:
> Looks like we either have to grab the small emote for some of the 3rd party ones or limit the size of the emotes when displayed. Screenshot_20220906-133502

Fixed in https://github.com/muxable/rtchat/pull/679

That looks hilarious though maybe we should leave it as an easter egg

PR #723, opened 2022-09-14
fix: add address to one webhook instead of one webhook per address

chungwwei:


functions/src/alchemy_webhook.ts
```diff
@@ -42,9 +43,29 @@ type AddressNotification = {
...
+  const hmac = crypto.createHmac("sha256", signingKey); // Create a HMAC SHA256 hash using the signing key
+  hmac.update(body, "utf8"); // Update the token hash with the request body using utf8
+  const digest = hmac.digest("hex");
+  return signature === digest;
+}
+
 export const alchemyWebhook = functions.https.onRequest(async (req, res) => {
   const body = req.body;
+  const headerKey = req.headers["x-alchemy-signature"] as string;
+  const signingKey = "FROM ALCHEMY DASHBOARD"; // TDOD: replace with your signing key from dashboard
```
kevmo314, line 60: Can you ask Danny for this?

kevmo314: I think this also needs to be stored in Github secrets. See how we handle the Twitch webhook for a similar example.

functions/src/alchemy_webhook.ts
```diff
@@ -105,45 +115,42 @@ export const setRealTimeCashAddress = functions.https.onCall(
...
+    // TODO: get webhoookId and X-Alchemy-Token from dashboard
     const options = {
-      method: "POST",
+      method: "PATCH",
       headers: {
-        Accept: "application/json",
-        "X-Alchemy-Token": "this-is-a-token", // TODO: get token from alchemy
-        "Content-Type": "application/json",
+        accept: "application/json",
+        "X-Alchemy-Token": "qewrewqrewqr",
```
kevmo314, line 123: Same for this token.

functions/src/alchemy_webhook.ts
```diff
@@ -105,45 +115,42 @@ export const setRealTimeCashAddress = functions.https.onCall(
...
+        "content-type": "application/json",
       },
       body: JSON.stringify({
-        network: "ETH_MAINNET",
-        webhook_type: "ADDRESS_ACTIVITY",
-        webhook_url: "localhost/{userId}", // TODO: replace with real webhook url
+        addresses_to_add: ["0x9131F047B512fD9cc42afc7025e2e9Cb7E62eedF"],
+        webhook_id: "qwereqwrqewreqw",
```
kevmo314, line 128: What's this webhook id?

chungwwei: it's the webhook created from dashboard, i just have some random letters there.

functions/src/alchemy_webhook.ts
juaoose: Or maybe firebase/gcp secrets? 

`firebase functions:secrets:set SECRET_NAME`
https://firebase.google.com/docs/functions/config-env

functions/src/alchemy_webhook.ts
```diff
@@ -42,9 +43,29 @@ type AddressNotification = {
...
+  const hmac = crypto.createHmac("sha256", signingKey); // Create a HMAC SHA256 hash using the signing key
+  hmac.update(body, "utf8"); // Update the token hash with the request body using utf8
+  const digest = hmac.digest("hex");
+  return signature === digest;
+}
+
 export const alchemyWebhook = functions.https.onRequest(async (req, res) => {
   const body = req.body;
+  const headerKey = req.headers["x-alchemy-signature"] as string;
+  const signingKey = functions.config().alchemy.signingkey;
```
kevmo314, line 60: Can you update `.github/workflows/firebase-functions.yml` to also set this secret? Otherwise it'll get overridden on deploy.

functions/src/alchemy_webhook.ts
```diff
@@ -105,45 +115,42 @@ export const setRealTimeCashAddress = functions.https.onCall(
...
-        const res = response as webhookIdResponse;
-        const webhookId = res.id;
         await admin
           .firestore()
           .collection("realtimecash")
           .doc(userId)
           .set({
             address,
-            webhookId,
+            WEBHOOKID,
```
kevmo314, line 144: `webhookId: WEBHOOKID`

kevmo314 (commented):
Mostly looks good just some config changes

kevmo314 (approved):

kevmo314:
We should store firebase secrets on GitHub, check the workflow. Secrets
might get overwritten in a future workflow run. GCP is ok but we should
ideally keep all our secrets in one place and so far GitHub is a bit easier.

PR #726, opened 2022-09-17
chore: error-logging for address setter

chungwwei:
Can't write to db after invoking the function, the firebase rules are there.

Logging to see what failed.


juaoose (approved):

kevmo314:
I like this style of debugging 😎

PR #731, opened 2022-09-19
fix: lowercase for eth public addr

chungwwei:
eth address not case sensitive, and the webhook response `toAddress` is all lowercase


juaoose (approved):
does adress case matter?

chungwwei:
> does adress case matter?
https://github.com/muxable/rtchat/blob/12490e63da2384eab09a77d5dca167f47b46d362/functions/src/alchemy_webhook.ts#L81-L88

I need lowercase comparison.

PR #732, opened 2022-09-20
fix: incorrect profile picture

kevmo314:


juaoose (approved):
Shouldn't we use the profile picture for both? the followed channels and the queried ones

kevmo314:
> Shouldn't we use the profile picture for both? the followed channels and the queried ones

Yeah, twitch is inconsistent on what thumbnail_url refers to

PR #742, opened 2022-09-21
feat: render crypto-msg

chungwwei:


lib/components/chat_history/auxiliary/realtimecash_donation.dart
```diff
@@ -0,0 +1,33 @@
+import 'package:flutter/material.dart';
+import 'package:rtchat/components/chat_history/decorated_event.dart';
+import 'package:rtchat/models/messages/auxiliary/realtimecash.dart';
+
+class RealTimeCashDonationEventWidget extends StatelessWidget {
```
kevmo314, line 5: Lowercase t, same with the model. `RealtimeCash`

lib/models/messages/auxiliary/realtimecash.dart
```diff
@@ -0,0 +1,25 @@
+import 'package:rtchat/models/messages/message.dart';
+
+class SimpleRealTimeCashDonationEventModel extends MessageModel {
+  final String assetName;
```
kevmo314, line 4: Make this `final AssetImage asset;` and pass the `AssetImage` into the constructor instead of the string. This keeps the resolution logic in the model.

kevmo314 (changes requested):

lib/models/messages/auxiliary/realtimecash.dart
```diff
@@ -0,0 +1,30 @@
...
+      required this.value,
+      required this.hash,
+      required String messageId,
+      required DateTime timestamp})
+      : super(messageId: messageId, timestamp: timestamp);
+
+  static SimpleRealtimeCashDonationEventModel fromDocumentData(
+      String messageId, Map<String, dynamic> data) {
+    return SimpleRealtimeCashDonationEventModel(
+        assetName: data['activity']['asset'] ?? "UNKNOWN",
```
kevmo314, line 22:
```suggestion
  final double value;
  final String hash;

  const SimpleRealtimeCashDonationEventModel(
      {required this.image,
      required this.value,
      required this.hash,
      required String messageId,
      required DateTime timestamp})
      : super(messageId: messageId, timestamp: timestamp);

  static SimpleRealtimeCashDonationEventModel fromDocumentData(
      String messageId, Map<String, dynamic> data) {
    return SimpleRealtimeCashDonationEventModel(
```

`assetName` can be removed since it isn't used anywhere I think?

kevmo314 (approved):

lib/models/messages/auxiliary/realtimecash.dart
kevmo314: Ohh, I see you're using `assetName` to mean currency. I would rename it to `currency` but up to you.

PR #745, opened 2022-09-21
feat: listen on eth main-net

chungwwei:


functions/src/alchemy_webhook.ts
```diff
@@ -54,6 +54,66 @@ function isValidSignatureForStringBody(
...
+) {
+  // look for userId that associated with this address
+  const addressDoc = await admin
+    .firestore()
+    .collection("realtimecash")
+    .where("address", "==", activity.toAddress)
+    .limit(1)
+    .get();
+
+  const userId = addressDoc.docs[0].id;
```
kevmo314, line 69: Log an error if `addressDoc.docs` is empty

kevmo314 (approved):
Not really part of this PR but just noticed one comment, otherwise looks good

PR #747, opened 2022-09-21
feat: add automatic translation

kevmo314:
The feature is currently hidden but this implements everything else necessary for automatic translation.


kevmo314:
I found a more clever way to do this.

PR #763, opened 2022-09-28
fix: add donor and msg to database, add those fields to model

chungwwei:


lib/components/chat_history/auxiliary/realtimecash_donation.dart
```diff
@@ -17,11 +17,13 @@ class RealtimeCashDonationEventWidget extends StatelessWidget {
...
             children: [
-              const TextSpan(text: "You received "),
+              TextSpan(text: "${model.donor} donated "),
               TextSpan(
-                  text: "${model.value.toString()} ${model.currency}.",
+                  text: "${model.value.toString()} ${model.currency}. ",
                   style: boldStyle),
-              const TextSpan(text: " Your transaction hash: "),
+              if (model.message.isNotEmpty)
+                TextSpan(text: "message: ${model.message}"),
```
kevmo314, line 25: Render the message in italics below the donation notification message: https://github.com/muxable/rtchat/blob/c16ceff81d06e99a32f400174911c48cda6064a8/lib/components/chat_history/auxiliary/streamlabs.dart#L27-L29

lib/components/chat_history/auxiliary/realtimecash_donation.dart
```diff
@@ -17,11 +17,13 @@ class RealtimeCashDonationEventWidget extends StatelessWidget {
...
+              const TextSpan(text: " Hash: "),
```
kevmo314, line 26: We can probably omit the hash. It's not really useful information. Maybe a useful way to use it would be long press opens up etherscan/polygonscan?

lib/models/messages/auxiliary/realtimecash.dart
```diff
@@ -6,12 +6,16 @@ class SimpleRealtimeCashDonationEventModel extends MessageModel {
   final String currency;
   final double value;
   final String hash;
+  final String donor;
```
kevmo314, line 9: Make these fields nullable instead of having a default value. It will make translations a bit easier too.

kevmo314 (changes requested):

lib/components/chat_history/auxiliary/realtimecash_donation.dart
kevmo314: This can be done in a separate PR, but I think we should remove the hash.

kevmo314 (approved):

PR #768, opened 2022-09-29
fix: guard null value in barcode scan

chungwwei:


kevmo314 (approved):

lib/screens/settings/donations.dart
```diff
@@ -80,11 +80,14 @@ class DonationsScreen extends StatelessWidget {
...
-                                                                .rawValue!
-                                                                .toLowerCase());
+                                                    if (barcode.rawValue !=
+                                                        null) {
+                                                      DonationsAdapter.instance
+                                                          .setRealtimeCashAddress(
+                                                              address: barcode
+                                                                  .rawValue!
+                                                                  .toLowerCase());
+                                                    }
```
kevmo314, line 90:
```suggestion
                                                    final value = barcode.rawValue;
                                                    if (value != null) {
                                                      DonationsAdapter.instance
                                                          .setRealtimeCashAddress(
                                                              address: value.toLowerCase());
                                                    }
```

We can avoid the non-null assertion. Might need a dart fmt.

PR #778, opened 2022-10-03
fix: fix parsing error in realtimecash dono model

chungwwei:


lib/models/messages/auxiliary/realtimecash.dart
```diff
@@ -24,7 +24,7 @@ class SimpleRealtimeCashDonationEventModel extends MessageModel {
       String messageId, Map<String, dynamic> data) {
     return SimpleRealtimeCashDonationEventModel(
         currency: data['activity']['asset'] ?? "UNKNOWN",
-        value: data['activity']['value'],
+        value: double.parse(data['activity']['value'].toString()),
```
kevmo314, line 27: Does `toDouble()` work?

kevmo314 (approved):

lib/models/messages/auxiliary/realtimecash.dart
chungwwei: i think same results

PR #809, opened 2022-11-02
fix: message input overflow when divider is drag too far

### LINKED ISSUE #805: [autocomplete] command autocomplete ui seems odd
image

Looks like the input field also gets pushed off screen?

chungwwei:
Closes #805

using `Stack` with `Positioned`, so that `InputField` is always on top of the keyboard

image

### DIFF
--- a/lib/components/chat_panel.dart
+++ b/lib/components/chat_panel.dart
@@ -176,7 +176,9 @@ class _ScrollToBottomWidget extends StatelessWidget {
   Widget build(BuildContext context) {
     return AnimatedPositioned(
       duration: const Duration(milliseconds: 300),
-      bottom: show ? 16 : -72,
+      bottom: show
+          ? 72 // 72 consideration for the InputField, since InputField is on top of the Stack.
+          : -72,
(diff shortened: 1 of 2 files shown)


chungwwei:
since the message `InputField` is last of the `Stack`, the bottommost message is covered up by the `InputField` entirely or partially.

kevmo314:
> since the message `InputField` is last of the `Stack`, the bottommost message is covered up by the `InputField` entirely or partially.
> 
> 

You mean even when the keyboard is not expanded?

chungwwei:
> > since the message `InputField` is last of the `Stack`, the bottommost message is covered up by the `InputField` entirely or partially.
> 
> You mean even when the keyboard is not expanded?

yes.

image

kevmo314:
> > > since the message `InputField` is last of the `Stack`, the bottommost message is covered up by the `InputField` entirely or partially.
> > 
> > 
> > You mean even when the keyboard is not expanded?
> 
> yes.
> 
> image

We should probably fix that?

PR #830, opened 2022-12-10
fix: properly parse alternate emote format

juaoose:


kevmo314:
This might've been my mistake (in other words doesn't actually happen in production) but I think it's ok to merge since it makes the parsing a little more resilient anyways.

kevmo314 (approved):

PR #831, opened 2022-12-10
feat: improve subscription event

kevmo314:
Screenshot_20221210-012430


SputNikPlop:
Twitch current adds a flair for different tier subscriptions with Tier 3 #ffea00 and Tier 2 #d2d2e6 colored flairs. Would it be ok to do different maybe colored text boxes to more easily read the tier level?

PR #841, opened 2022-12-13
fix: ios webview muting

kevmo314:


SputNikPlop (approved):
works fine on the smiluator

PR #860, opened 2022-12-26
fix: small format changes

SputNikPlop:


kevmo314 (approved):
Are these from dart fmt?

SputNikPlop:
> Are these from dart fmt?

YEP

PR #880, opened 2023-01-16
chore: add message dialog localization

juaoose:


kevmo314 (approved):
Ah don't worry too much about getting the remaining strings switched over, I got someone to do the work one-off for us.

PR #900, opened 2023-01-24
feat: wired up shoutout

chungwwei:
I'm not getting the shoutout events in the db, there's probably a bug somewhere.


lib/components/chat_history/twitch/shoutout_event.dart
```diff
@@ -16,12 +21,50 @@ class TwitchShoutoutEventWidget extends StatelessWidget {
                   text: "Shoutout was given to ",
```
kevmo314, line 21: Can we match the Twitch text for shoutouts? @SputNikPlop do you know what it is?

kevmo314 (approved):

lib/components/chat_history/twitch/shoutout_event.dart
SputNikPlop: > Can we match the Twitch text for shoutouts? @SputNikPlop do you know what it is?

It looks something like this
Screen Shot 2023-01-27 at 4 16 43 PM

kevmo314 (approved):

SputNikPlop (approved):

PR #907, opened 2023-01-27
chore: removed incorrectly parsed data in poll

chungwwei:


SputNikPlop:
@kevmo314 Some polls do have this data for those we should display it no when we get the data?

chungwwei:
bitvotes is never used, and is always set to `false` in `is_enabled`
https://dev.twitch.tv/docs/eventsub/eventsub-reference/#channel-prediction-begin-event 
https://dev.twitch.tv/docs/eventsub/eventsub-reference/#choices
so we should remove that.

channel points is used if enabled, should render if it its enabled. To save space, we can do text ellipsis on title and have the channel points votes to the right of tilte if channel points votes is enabled.

`votes` already account for the one vote per person and any additional votes with channel points
image

kevmo314 (approved):

SputNikPlop (approved):

PR #908, opened 2023-01-27
fix: twitch and app mismatch

SputNikPlop:
Screen Shot 2023-01-27 at 1 19 45 AM
Makes third party match first party behavior


KeN095 (approved):

SputNikPlop:
The text from twitch reads now "They've gifted a total of ${model.cumulativeTotal} in the channel", I currently like the the current way. Another thing to note it seems to still count sub months rather than exact gift amounts.

kevmo314:
> The text from twitch reads now "They've gifted a total of ${model.cumulativeTotal} in the channel", I currently like the the current way. Another thing to note it seems to still count sub months rather than exact gift amounts.

Yeah agreed, we should bias towards being more concise because we're on smaller screens. Interesting that Twitch changed it.

PR #909, opened 2023-01-28
fix: android branding

SputNikPlop:


cswills:
I tested this and the Muxable logo is stretched. I think the branding images need to have an aspect ratio of 5:2

Screenshot 2023-01-27 at 9 23 40 PM

SputNikPlop:
> I tested this and the Muxable logo is stretched. I think the branding images need to have an aspect ratio of 5:2
> 
> Screenshot 2023-01-27 at 9 23 40 PM

Ah thanks for the catch I will ask about getting one done.

cswills (approved):

PR #913, opened 2023-01-28
fix: consistent emote picker theming

cswills:
The tab indicator in the emote picker uses the default material theme accent color. Also, the sticky header color isn't consistent between light and dark theme.


SputNikPlop:
Just tested and its working, awesome work.

SputNikPlop (approved):

PR #914, opened 2023-01-28
fix: older android compatibility

SputNikPlop:
#910


SputNikPlop:
Will make another after #909 gets merged

PR #927, opened 2023-02-01
feat: allow tts to play in the background on ios

juaoose:
These apis seem to change between iOS versions which is why I excluded some of the common options, tested on 16.0


SputNikPlop (approved):
Looks good, I think when more people upgrade their iOS then we can tweak more options slowly. Worked ok even on iPhone 5s, tested on iPhone 13 iOS 15/16 worked fine as well.

PR #933, opened 2023-02-03
chore: replaced textlink shoutout with eventsub shoutout in inbound raid

### LINKED ISSUE #918: [chat_history] add "shoutout" button to raiding message
Similar to "join" for the outgoing raids, let's add a "shoutout" button to incoming raids that sends the shoutout command.

chungwwei:
Closes #918

its already implemented but ig we can trigger the eventsub with the text `/shoutout <streamer>`

### DIFF
--- a/lib/components/chat_history/twitch/raid_event.dart
+++ b/lib/components/chat_history/twitch/raid_event.dart
@@ -1,12 +1,10 @@
 import 'package:flutter/material.dart';
 import 'package:intl/intl.dart';
-import 'package:provider/provider.dart';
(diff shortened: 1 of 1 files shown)


lib/components/chat_history/twitch/raid_event.dart
```diff
@@ -51,6 +51,8 @@ class TwitchRaidEventWidget extends StatelessWidget {
               onTap: () {
                 ActionsAdapter.instance
```
kevmo314, line 52: Let's remove this and only run the shoutout command

kevmo314 (approved):

PR #938, opened 2023-02-05
fix: const simplification

SputNikPlop:


cswills (changes requested):
`Row` doesn't have a const constructor. Check flutter analyzer.

lib/screens/settings/dismissible_delete_background.dart
```diff
@@ -7,10 +7,10 @@ class DismissibleDeleteBackground extends StatelessWidget {
   Widget build(BuildContext context) {
     return Container(
         color: Colors.red,
-        child: Padding(
-            padding: const EdgeInsets.symmetric(horizontal: 16),
+        child: const Padding(
```
cswills, line 10: I don't think it's possible to further simplify these const statements. You can't use the const keyword on `Padding` because its child, `Row`, doesn't have a const constructor.

SputNikPlop: I think you are right, I get it wrong sometimes so I will close this pr.

PR #947, opened 2023-02-11
chore: eventsub shoutout, from beta to version 1

chungwwei:


functions/src/eventsub.ts
```diff
@@ -53,10 +53,7 @@ function createEventsub(token: string, type: string, twitchUserId: string) {
       moderator_user_id: twitchUserId,
     };
   }
-  const version =
-    type === "channel.shoutout.create" || type === "channel.shoutout.receive"
-      ? "beta"
-      : "1";
+  const version = "1";
```
juaoose, line 56: can we inline version to the body?

juaoose (approved):

PR #949, opened 2023-02-12
fix: button visibility

SputNikPlop:
My idea to move the button up as it is more of an app setting than a system setting, should also help people find it.

### DIFF
--- a/lib/screens/settings/settings.dart
+++ b/lib/screens/settings/settings.dart
@@ -82,6 +82,14 @@ class _SettingsScreenState extends State<SettingsScreen> {
               Navigator.pushNamed(context, "/settings/third-party");
             },
           ),
+          SwitchListTile.adaptive(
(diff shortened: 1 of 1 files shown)


lib/screens/settings/settings.dart
```diff
@@ -82,6 +82,14 @@ class _SettingsScreenState extends State<SettingsScreen> {
               Navigator.pushNamed(context, "/settings/third-party");
             },
           ),
+          SwitchListTile.adaptive(
+            title: const Text('Show viewer and follower count'),
+            subtitle: const Text('Disable/Enable across the app'),
```
kevmo314, line 87: I don't think the subtitle is necessary.

kevmo314 (approved):

lib/screens/settings/settings.dart
SputNikPlop: Removed.

PR #961, opened 2023-02-21
fix: add german translation file

SputNikPlop:
This is the German translation file.


SputNikPlop:
No longer needed this was a template see #970

PR #962, opened 2023-02-21
feat: french translation

SputNikPlop:
This is the French translation. Work in progress so it will be a draft then will be converted to a pull request.


SputNikPlop:
See #972

PR #1011, opened 2023-04-21
feat: bengali language

SputNikPlop:


juaoose:
Maybe it would be better to not include the strings for which we don't have the bengali translation instead of spanish?

SputNikPlop:
> Maybe it would be better to not include the strings for which we don't have the bengali translation instead of spanish?

I think we should add a template, mastema who did this I think used the spanish one at the time as a template.

kevmo314 (approved):

kevmo314:
> > Maybe it would be better to not include the strings for which we don't have the bengali translation instead of spanish?
> 
> I think we should add a template, mastema who did this I think used the spanish one at the time as a template.

It's ok to leave out strings without translations. The English file should be used as the template, remove the strings with @. It's not super intuitive though but not sure if there's a better solution.

PR #1013, opened 2023-04-23
fix: template for translations

SputNikPlop:
@juaoose @kevmo314 this is the template file, I don't know how to properly include it though


kevmo314:
I don't think we should use an explicit template file, it will be a pain to keep in sync with english

SputNikPlop:
> I don't think we should use an explicit template file, it will be a pain to keep in sync with english

What or which way would be best for us to go about making it simple and easy for translators to contribute? Most of our contributors so far use the app but are unfamiliar with the codebase.

kevmo314:
> > I don't think we should use an explicit template file, it will be a pain to keep in sync with english
> 
> What or which way would be best for us to go about making it simple and easy for translators to contribute? Most of our contributors so far use the app but are unfamiliar with the codebase.

We probably can setup a small website that generates the translation table. This would make it easier for people to contribute and I don't think it would be too difficult. I'll add it onto the pile lol

SputNikPlop:
> We probably can setup a small website that generates the translation table. This would make it easier for people to contribute and I don't think it would be too difficult. I'll add it onto the pile lol

Alright added #1017 and will close this issue

PR #1022, opened 2023-05-05
feat: chinese translation

SputNikPlop:


lib/l10n/app_cn.arb
```diff
@@ -0,0 +1,112 @@
+{
```
kevmo314, line 1: File name should be `app_zh.arb`

juaoose (approved):

SputNikPlop:
This PR will 2023-03-09 as I review some of the phrases.

PR #1046, opened 2023-06-27
feat: reenable persistence

kevmo314:


SputNikPlop (approved):
LGTM

I ended up changing removing the code because as it is it was more efficient. Tested with this fix https://github.com/firebase/flutterfire/pull/11174 which ended up getting pushed into and bumped it looks fine to me right now. The current way saves about 10MB of RAM whereas we would have added extra overhead.

PR #1059, opened 2023-07-25
fix: upgrade dependencies

SputNikPlop:


pubspec.yaml
```diff
@@ -12,44 +12,44 @@ dependencies:
...
+  cloud_firestore: ^4.8.3
+  cloud_functions: ^4.3.4
+  crypto: ^3.0.3
   cupertino_icons: ^1.0.5
+  firebase_analytics: ^10.4.4
+  firebase_auth: ^4.7.1
+  firebase_core: ^2.15.0
+  firebase_crashlytics: ^3.3.4
+  firebase_database: ^10.2.4
+  flutter_custom_tabs: ^1.0.4
```
SputNikPlop, line 27: Something appears to be weird when upgrading this keeping at 1.0.4 for now

PR #1072, opened 2023-09-16
Revert "feat: use twitch irc for sending messages"

kevmo314:
Reverts muxable/rtchat#1067


SputNikPlop (approved):
LGTM Sadge cuz twitch

PR #1077, opened 2023-09-29
fix: upgrade webview

SputNikPlop:


SputNikPlop:
> You tested it works and all yeah?

Yes, I am still playing around and testing stuff

PR #1084, opened 2023-10-17
feat: create the groundwork for a native android tts channel

jlw4049:


lib/screens/home.dart
```diff
@@ -237,6 +238,9 @@ class _HomeScreenState extends State<HomeScreen> with TickerProviderStateMixin {
                             : Icons.voice_over_off),
                         tooltip: AppLocalizations.of(context)!.textToSpeech,
                         onPressed: () {
+                          // TODO: need to figure out how to actually
+                          // send the text to this plugin, but manual is working
+                          // TextToSpeechPlugin.speak("Testing this code!");
```
jlw4049, line 243: This is working however, I'm not 100% how you want the text fed to the plugin

ps-kwang: This is fine for now.

Issue #904, opened 2023-01-26
[chat] qr codes


SputNikPlop:
https://clips.twitch.tv/WealthySwissPresidentNerfRedBlaster-1uszjVbudlI7fQpi

ryansng:
I remember writing code for this exact use-case with real time chat logo in the QR code but idea was gutted ):

ryansng:
![image](https://user-images.githubusercontent.com/86465073/215061454-5b3d1d88-1358-4483-bb73-e0a2dd11825c.png)
![image](https://user-images.githubusercontent.com/86465073/215061466-79593c63-66a6-46e1-9d6c-4671f81fd6cc.png)

SputNikPlop:
> I could take this up again at some point if it's not urgent

Feel free there is no rush at all on this at all.

SputNikPlop:
![IMG_1465](https://github.com/muxable/rtchat/assets/100245448/8842175e-436d-4628-8970-073da83382a7)
![IMG_1466](https://github.com/muxable/rtchat/assets/100245448/74b1aa4d-d6de-4e98-a08c-2da6c7cf312c)
@kevmo314 Which do you like more I am leaning towards the second as it look more like a card

PR #1096, opened 2023-10-28
feat: qr codes

SputNikPlop:
This is my partial attempt on how to get #904 implemented. Any feedback and help is welcome.

### DIFF
--- a/lib/l10n/app_bn.arb
+++ b/lib/l10n/app_bn.arb
@@ -68,6 +68,8 @@
   "textToSpeechSubtitle": "টেক্সট টু স্পিচ সেটিংস পরিবর্তন করুন",
   "events": "ইভেন্টস",
   "eventsSubtitle": "টুইচ ইভেন্ট কনফিগার করুন",
+  "qr": "QR",
+  "qrSubtitle": "QR কনফিগার করুন",
   "thirdPartyServices": "থার্ড পার্টি সার্ভিসেস",
   "thirdPartyServicesSubtitle": "থার্ড পার্টি সার্ভিসেসের সাথে কানেক্ট করুন",
   "followingEvent": "<bold>{displayName}</bold> আপনাকে অনুসরন করছে",
--- a/lib/l10n/app_zh_Hant.arb
+++ b/lib/l10n/app_zh_Hant.arb
@@ -68,6 +68,8 @@
   "textToSpeechSubtitle": "變更文字轉語音設定",
   "events": "事件",
   "eventsSubtitle": "設定 Twitch 事件",
+  "qr": "二维码",
+  "qrSubtitle": "設定  二维码",
   "thirdPartyServices": "第三方服務",
   "thirdPartyServicesSubtitle": "連結至第三方服務",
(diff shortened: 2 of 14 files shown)


lib/l10n/app_bn.arb
```diff
@@ -68,6 +68,8 @@
   "textToSpeechSubtitle": "টেক্সট টু স্পিচ সেটিংস পরিবর্তন করুন",
   "events": "ইভেন্টস",
   "eventsSubtitle": "টুইচ ইভেন্ট কনফিগার করুন",
+  "qr": "QR",
+  "qrSubtitle": "QR কনফিগার করুন",
```
SputNikPlop, line 72: This should be reviewed by mastema

lib/l10n/app_zh_Hant.arb
```diff
@@ -68,6 +68,8 @@
   "textToSpeechSubtitle": "變更文字轉語音設定",
   "events": "事件",
   "eventsSubtitle": "設定 Twitch 事件",
+  "qr": "二维码",
+  "qrSubtitle": "設定  二维码",
```
SputNikPlop, line 72: Should double check this with native speakers.

lib/screens/settings/qr.dart
```diff
@@ -0,0 +1,115 @@
...
+              Slider.adaptive(
+                value: qrModel.version,
+                min: 3,
+                max: 16,
+                divisions: 13,
+                label: "${qrModel.version.toInt()}",
+                onChanged: (value) {
+                  qrModel.size = value;
+                },
+              ),
```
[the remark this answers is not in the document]
SputNikPlop: This should just change the pattern of the QR code, pretty sure it is leftover from when I was prototyping.

lib/screens/settings/qr.dart
```diff
@@ -0,0 +1,115 @@
...
+              SwitchListTile.adaptive(
+                title: const Text('Use  profile image in the qr'),
+                value: qrModel.useProfile,
+                onChanged: (value) {
+                  qrModel.toggleProfileImage();
+                },
+              ),
```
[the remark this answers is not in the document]
SputNikPlop: The theme changes on tap. What should be the options that are allowed to be configured?

lib/models/qr_code.dart
```diff
@@ -0,0 +1,215 @@
...
+          const Color(0x00fdcbf1),
+          const Color(0x00e6dee9)
+        ],
+        [0.0, 0.01, 1.0],
+        tileMode,
+      );
+
+  static Gradient winterNeva({TileMode tileMode = TileMode.clamp}) => linear(
+        "Winter Neva",
+        pi / 6,
```
SputNikPlop, line 176: @kevmo314 Let me know if I did this right.

SputNikPlop:
@kevmo314 Ok good points made, also that makes more sense since on the issue I had to drop the modal sheet because I could only get that to work with iOS 13

SputNikPlop:
Screen Shot 2023-11-20 at 1 39 29 PM
@kevmo314 what do you think about moving the /username part outside the card part?

PR #1107, opened 2023-11-05
fix: broken 7tv emotes

SputNikPlop:
The global 7tv emotes appear to be broken because the api has been deprecated.


functions/src/emotes.ts
```diff
@@ -209,7 +209,7 @@ export async function get7TVEmotes(channelId: string): Promise<Emote[]> {
       }))
     )
     .catch(() => []);
-  const local = fetch(`https://api.7tv.app/v2/users/${channelId}/emotes`)
+  const local = fetch(`https://7tv.io/v3/users/${channelId}/emotes`)
```
SputNikPlop, line 212: Unsure if this is right based on https://7tv.io/docs

PR #1122, opened 2023-11-28
fix: add tts unit test

### LINKED ISSUE #1102: Add unit tests for the circular buffer

jlw4049:
fixes #1102


test/models/tts_test.dart
```diff
@@ -0,0 +1,54 @@
...
+
+    test('Queue starts empty', () {
+      expect(ttsQueue.isEmpty, isTrue);
+    });
+
+    test('Speak adds elements to the queue', () {
+      ttsQueue.speak('1', 'First message');
+      ttsQueue.speak('2', 'Second message');
+      expect(ttsQueue.isEmpty, isFalse);
+      expect(ttsQueue.length, equals(2));
```
kevmo314, line 24: Shouldn't this be 1 because the first invocation gets sent to the channel?

test/models/tts_test.dart
```diff
@@ -0,0 +1,54 @@
...
+
+      ttsQueue.delete('1');
+
+      expect(ttsQueue.length, equals(1));
+      expect(ttsQueue.peek(), equals({'id': '2', 'text': 'Second message'}));
+    });
+  });
+}
+
+class MockMethodChannel extends Mock implements MethodChannel {}
```
kevmo314, line 54: Can you also add a test that resolving speak from the mocked channel lines up the next message?

pubspec.yaml
```diff
@@ -53,6 +53,7 @@ dependencies:
   webview_flutter_wkwebview: ^3.9.1
 
 dev_dependencies:
+  mockito: ^5.0.0
```
SputNikPlop, line 56: Possible to use a newer version?

pubspec.yaml
```diff
@@ -53,6 +53,7 @@ dependencies:
   webview_flutter_wkwebview: ^3.9.1
 
 dev_dependencies:
+  mockito: ^5.7.0
```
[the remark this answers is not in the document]
jlw4049: Ah yes, I was looking at the official package not the flutter. Got it!

test/models/tts_test.dart
jlw4049: Let me know if the changes resolved it. The only way I could figure out how to get it to only have 1 in the queue is to actually "speak" the message by using speakNext(). If this is not what needed to be done let me know!

test/models/tts_test.dart
jlw4049: Was a bit tougher than the others, but I think I got it

test/models/tts_test.dart
```diff
@@ -0,0 +1,77 @@
...
+      ttsQueue.speak('2', 'Second message');
+
+      when(mockedTTSQueue.speak(any, any)).thenAnswer((_) async {});
+
+      await ttsQueue.speakNext();
+      expect(ttsQueue.length, equals(1));
+      expect(ttsQueue.peek(), equals({'id': '2', 'text': 'Second message'}));
+
+      await ttsQueue.speakNext();
+      expect(ttsQueue.isEmpty, isTrue);
```
kevmo314, line 74:
```suggestion
      var callCount = 0;
      when(mockedTTSQueue.speak(any, any)).thenAnswer((_) async {
        callCount++;
        await Future.delayed(seconds: 1);  // prevent race conditions
      });

      expect(callCount, 0);
      const promise1 = ttsQueue.speak('1', 'First message');
      const promise2 = ttsQueue.speak('2', 'Second message');
      
      expect(ttsQueue.length, equals(1));  // pop off immediately
      await promise1;
      expect(callCount, 1);
      expect(ttsQueue.length, equals(1));  // resolve before the next tts speak command has finished
      await promise2;
      expect(callCount, 2);
      expect(ttsQueue.length, equals(0)); 
```

Nice, this is on the right track. We can avoid the `speak`/`speakNext()` split by counting in the mock. I think the reason this is important is there is a bug in the code right now:

```
const run1 = speak(1);
const run2 = speak(2);

await run1; // this will wait until run2 finishes
await run2; // this will return immediately
```

but what we want is:

```
const run1 = speak(1);
const run2 = speak(2);

await run1; // this will wait until run1 finishes
await run2; // this will wait until run2 finishes
```

test/models/tts_test.dart
kevmo314: These are on the right track. Nice work figuring this out, check out the other comment. I think there's a bug in the current implementation.

kevmo314:
Tests seem to fail on CI? https://github.com/muxable/rtchat/actions/runs/7051914298/job/19195849795?pr=1122

jlw4049:
> Tests seem to fail on CI? https://github.com/muxable/rtchat/actions/runs/7051914298/job/19195849795?pr=1122

Yes, I'm going to try to sort that last test without the use of mockito today so that we're passing. The last push I did was broken.

PR #1123, opened 2023-11-29
fix: viewer list error message

SputNikPlop:
This is my hack to showing an error, not sure the best way to address the problem so I'll iterate on this.


lib/components/drawer/end_drawer.dart
```diff
@@ -22,13 +158,48 @@ class EndDrawerWidgetState extends State<EndDrawerWidget> {
...
+            Fluttertoast.showToast(
+              msg: "Loading data, please wait...",
+              toastLength: Toast.LENGTH_SHORT,
+              gravity: ToastGravity.BOTTOM,
+              timeInSecForIosWeb: 1,
+              backgroundColor: Colors.blue,
+              textColor: Colors.white,
+              fontSize: 10.0,
+            );
+            return Container();
```
kevmo314, line 174: You can use the `ScaffoldMessenger` for toasts

```
ScaffoldMessenger.of(context).showSnackBar(const SnackBar(
     content: Text("Sending Message"),
));
```

But this is a little tricky to do correctly because the builder is supposed to be side-effect free. What you'll want to do instead is not use `FutureBuilder` and do the following things:

1. Update `EndDrawerWidget` to accept an `onError` callback.
2. Update the caller to, on error, close the drawer and show the snackbar.
3. Create a `Viewers? _viewers` state parameter like `String _search` right now
4. Add an `initState() {} ` function that calls `ChatStateAdapter.instance.getViewers()`
5. If the response is successful, `setState(() { _viewers = response; })`
6. If the response fails, call the `widget.onError()` callback.

kevmo314 (changes requested):

kevmo314:
Nice work, this looks like some good practice.

PR #1126, opened 2023-12-07
fix: native tts on ios

SputNikPlop:


kevmo314 (commented):
This is not called, but for the tts project right?

ios/Runner/AppDelegate.swift
```diff
@@ -11,6 +12,41 @@ import WebKit
...
+            let synthesizer = AVSpeechSynthesizer()
+
+            switch call.method {
+            case "speak":
+                guard let text = args?["text"] as? String, !text.isEmpty else {
+                    return result(FlutterError.init(code: "INVALID_ARGUMENT", message: "Text is empty or null", details: nil))
+                }
+                
+                let utterance = AVSpeechUtterance(string: text)
+                synthesizer.speak(utterance)
```
kevmo314, line 28:
```suggestion
                synthesizer.speak(utterance)
                result(Bool(true))
```

Missing a result

ios/Runner/AppDelegate.swift
```diff
@@ -11,6 +12,41 @@ import WebKit
...
         let controller: FlutterViewController = window?.rootViewController as! FlutterViewController
+        let ttsChannel = FlutterMethodChannel(name: "tts_plugin", binaryMessenger: controller.binaryMessenger)
+        ttsChannel.setMethodCallHandler {
+            (call: FlutterMethodCall, result: @escaping FlutterResult) in
+            let args = call.arguments as? [String: Any]
+            let synthesizer = AVSpeechSynthesizer()
+
+            switch call.method {
+            case "speak":
+                guard let text = args?["text"] as? String, !text.isEmpty else {
```
kevmo314, line 23: Empty text is ok, actually. If the text is empty return `result(Bool(true))` immediately.

kevmo314 (approved):

PR #1133, opened 2023-12-21
fix: notification to channel language

### LINKED ISSUE #919: [chat_history] switch notification messages to use the language of the channel
The language of the follow, sub, etc notifications should match the language of the channel if we have translations in that language. This is found in the channel's metadata.

SputNikPlop:
This aims to resolve #919 
I need this looked over first also probably want to update all the translation messages in this PR.


SputNikPlop:
> What does twitch translate the tiers as? It should match the labeling they use since they are more like proper nouns.

muxfd subscribed at Tier 1. They've subscribed for 10 months! or muxfd subscribed at Tier 1. They've subscribed for 10 months, currently on a 10 month streak! Should probably also think about exposing the streak amount maybe as well.

lib/components/chat_history/twitch/subscription_event.dart
```diff
@@ -52,12 +56,13 @@ class TwitchSubscriptionGiftEventWidget extends StatelessWidget {
                 style: Theme.of(context).textTheme.titleSmall),
             TextSpan(
                 text:
-                    " gifted ${model.total} Tier ${model.tier.replaceAll("000", "")} "),
+                    " ${localizations!.gifted} ${model.total} ${localizations.tier} ${model.tier.replaceAll("000", "")} "),
             TextSpan(
-                text: model.total > 1 ? "subscriptions. " : "subscription. "),
```
SputNikPlop, line 57: Do you know what this part of the logic was I don't remember off the top of my head, would like to make the translation part more descriptive for our translators.

SputNikPlop:
The subscription message I will investigate into as a separate PR.

lib/components/chat_history/twitch/subscription_event.dart
```diff
@@ -28,6 +28,7 @@ class TwitchSubscriptionEventWidget extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final localizations = AppLocalizations.of(context);
```
kevmo314, line 31: These are being reported as unused?

SputNikPlop: I think this was actually resolved in https://github.com/muxable/rtchat/pull/1270 so I don't need this anymore?

SputNikPlop:
Superceeded by #1335

PR #1135, opened 2023-12-23
fix: qr visibility improvements

### LINKED ISSUE #1120: QR code improvements
Screen Shot 2023-11-26 at 8 08 20 PM

These are small improvements that will make QR codes better in the app.

- Make the profile picture round as @kevmo314 mentioned in #1096 
- Make the dots black for better camera visibility

SputNikPlop:
This should fix #1120 the only thing I have left on this is to make the square a circle. @kevmo314 Whats the best way you think to do this without modifying the package?


SputNikPlop:
> > This should fix #1120 the only thing I have left on this is to make the square a circle. @kevmo314 Whats the best way you think to do this without modifying the package?
> 
> Screenshot of what it looks like right now?

Screen Shot 2023-12-23 at 3 24 33 AM

we should get the middle part to look something like

Screen Shot 2023-12-23 at 3 23 23 AM

PR #1139, opened 2023-12-27
fix: circular avatar

SputNikPlop:


lib/screens/settings/qr.dart
```diff
@@ -32,20 +32,27 @@ class QRDisplay extends StatelessWidget {
...
+                          color: Colors.black,
+                        ),
+                        dataModuleStyle: const QrDataModuleStyle(
+                          dataModuleShape: QrDataModuleShape.square,
+                          color: Colors.black,
+                        ),
+                        // Remove the embeddedImage property
+                      ),
+                      CircleAvatar(
+                        radius: 35,
```
SputNikPlop, line 52: Is this a fine amount? I picked 35 as usually the images we use are 70x70
Screen Shot 2023-12-27 at 2 48 15 AM

PR #1146, opened 2023-12-28
fix: incorrect empty property

kevmo314:


SputNikPlop (approved):
derp forgot about this

PR #1147, opened 2023-12-28
Isolate background service

jlw4049:
fixes #1104

PR #1150, opened 2023-12-29
fix: get background task working on ios

SputNikPlop:
I couldn't get #1147 working on my end kinda got it working with this hack but it is far from perfect.


android/app/src/main/AndroidManifest.xml
```diff
@@ -75,5 +75,9 @@
             android:name=".AudioService"
             android:enabled="true"
             android:foregroundServiceType="mediaPlayback" />
+
+        <service
+            android:name="id.flutter.flutter_background_service.BackgroundService"
```
SputNikPlop, line 80: This task of background service might run into some issues on some models of phone because of battery optimizations, possibly something to do with battery optimizations https://stackoverflow.com/questions/39256501/check-if-battery-optimization-is-enabled-or-not-for-an-app/54982071#54982071

android/build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.2.0'
+        classpath 'com.android.tools.build:gradle:7.4.2'
```
SputNikPlop, line 9: Will not work on any gradle below this version

ios/Runner/Info.plist
```diff
@@ -2,6 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+    <key>BGTaskSchedulerPermittedIdentifiers</key>
```
SputNikPlop, line 5: This only fixes things on iOS 13 plus apparently when running on my iPhone 5s it fails. I think we should move this to the AppDelegate though I haven't found a  way yet.

ios/Runner/Info.plist
```diff
@@ -2,6 +2,10 @@
...
+	<array>
+		<string>dev.flutter.background.refresh</string>
```
SputNikPlop, line 7: Maybe also make our own custom task identifier here because apple is notoriously picky for background things. After 15 minutes I hit a hiccup on the latest version of iOS 17 simulator.

lib/main.dart
```diff
@@ -56,12 +56,14 @@ import 'package:rtchat/screens/settings/tts/voices.dart';
...
 import 'package:rtchat/themes.dart';
 import 'package:shared_preferences/shared_preferences.dart';
+import 'package:rtchat/tts_isolate.dart' as ttsIsolate;
 
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
 
   await Firebase.initializeApp(options: DefaultFirebaseOptions.currentPlatform);
   await MobileAds.instance.initialize();
+  await ttsIsolate.initializeService();
```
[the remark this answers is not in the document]
SputNikPlop: Ok got it

SputNikPlop:
> I think this should be combined with @jlw4049's pr

I think so as well though I don't want to mess his PR up, also see my code review comments there are minor things that might need a look at

android/app/src/main/AndroidManifest.xml
SputNikPlop: Thinking about this and how the beacon app is maybe we should request it be turned off before the tts service starts

android/app/src/main/AndroidManifest.xml
```diff
@@ -75,5 +75,9 @@
...
+            android:foregroundServiceType="idkwhattoputhere"/>
```
SputNikPlop, line 81: Perhaps https://developer.android.com/about/versions/14/changes/fgs-types-required#remote-messaging

ios/Runner/Info.plist
SputNikPlop: Looking at this further I think I should remove it from the plist and just use https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/using_background_tasks_to_update_your_app

android/app/src/main/AndroidManifest.xml
kevmo314, line 81:
```suggestion
            android:foregroundServiceType="mediaPlayback"/>
```

https://developer.android.com/about/versions/14/changes/fgs-types-required#media

lib/main.dart
```diff
@@ -55,15 +55,18 @@ import 'package:rtchat/screens/settings/tts/languages.dart';
...
 
 void main() async {
   WidgetsFlutterBinding.ensureInitialized();
 
   await Firebase.initializeApp(options: DefaultFirebaseOptions.currentPlatform);
   await MobileAds.instance.initialize();
+  //wait ttsIsolate.initializeService();
 
-  final prefs = await SharedPreferences.getInstance();
+  final prefs = await StreamingSharedPreferences.instance;
```
kevmo314, line 69: This shouldn't be changed in the main isolate.

lib/main.dart
```diff
@@ -95,7 +98,7 @@ void main() async {
 }
 
 class App extends StatefulWidget {
-  final SharedPreferences prefs;
+  final StreamingSharedPreferences prefs;
```
kevmo314, line 101: Same here, this doens't need to be changed.

lib/main.dart
```diff
@@ -116,26 +119,27 @@ class _AppState extends State<App> {
...
             ..addListener(() {
               widget.prefs
                   .setString("activity_feed", jsonEncode(model.toJson()));
             });
         }),
         ChangeNotifierProvider(create: (context) {
-          final model = LayoutModel.fromJson(
-              jsonDecode(widget.prefs.getString("layout") ?? "{}"));
+          final model = LayoutModel.fromJson(jsonDecode(
+              widget.prefs.getString("layout", defaultValue: '').getValue()));
```
kevmo314, line 133: `jsonDecode()` will fail here, the `fromJson` call requires an empty object I think?

lib/main.dart
```diff
@@ -179,81 +186,102 @@ class _AppState extends State<App> {
...
+              ..addListener(() {
+                widget.prefs.setString('audio', jsonEncode(model.toJson()));
+              });
+          },
+          update: (context, userModel, model) {
+            model!.hostChannel = userModel.userChannel;
+            return model;
+          },
+          lazy: false,
+        ),
```
kevmo314, line 280: Dupe?

lib/screens/home.dart
```diff
@@ -243,13 +244,15 @@ class _HomeScreenState extends State<HomeScreen> with TickerProviderStateMixin {
...
                         // Test the isolate
-                        // if (ttsModel.enabled) {
-                        //   await Isolate.spawn(
-                        //       ttsIsolate.isolateMain, ReceivePort().sendPort);
-                        // }
+                        if (ttsModel.enabled) {
+                          FlutterBackgroundService().invoke("startTts");
+                          FlutterBackgroundService()
+                              .invoke('initSharedPreference');
+                        }
```
kevmo314, line 255: Let's not trigger this in prod yet.

lib/screens/home.dart
```diff
@@ -243,13 +244,15 @@ class _HomeScreenState extends State<HomeScreen> with TickerProviderStateMixin {
 
                         if (!ttsModel.enabled) {
                           await TextToSpeechPlugin.clear();
+                          FlutterBackgroundService().invoke("stopService");
```
kevmo314, line 247: Same here, this should be under a more strict flag.

lib/tts_isolate.dart
```diff
@@ -1,7 +1,73 @@
...
+import 'dart:ui';
+import 'package:flutter/material.dart';
+import 'package:flutter_background_service/flutter_background_service.dart';
+import 'package:streaming_shared_preferences/streaming_shared_preferences.dart';
 
 @pragma("vm:entry-point")
 void isolateMain(SendPort sendPort) {
-  print("Hello from the isolate");
-  sendPort.send("Isolate message");
+  initializeService(sendPort);
```
kevmo314, line 9: The service should be initialized on the main thread.

kevmo314 (changes requested):

lib/main.dart
SputNikPlop: Yeah

lib/screens/home.dart
SputNikPlop: Alright

lib/screens/home.dart
SputNikPlop: I kept this commented in otherwise the CI would fail.

android/app/src/main/AndroidManifest.xml
kevmo314: The service doesn't run unless explicitly requested and foregrounded.

ios/Runner/Info.plist
kevmo314: I don't think this entry is necessary, we don't need to update the app in the background only run code in the isolate.

android/app/src/main/AndroidManifest.xml
SputNikPlop: Ok then this isn't necessary

kevmo314: No, the service declaration is still necessary otherwise the server won't be able to start. The manifest doesn't cause the service to run.

ios/Runner/Info.plist
SputNikPlop: Don't I still have to add it for iOS https://pub.dev/documentation/flutter_background_service/latest/ the example of the plist is just for iOS 13+ though. Adding it I thought would resolve the background task and the other issue of it being killed in the background.

android/app/src/main/AndroidManifest.xml
SputNikPlop: I meant the battery optimization bit which I was worried about.

ios/Runner/Info.plist
SputNikPlop: @kevmo314 ok I updated it opting to use the AppDelegate.swift instead of the plist

ios/Runner/AppDelegate.swift
```diff
@@ -9,6 +10,8 @@ import WebKit
         _ application: UIApplication,
         didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?
     ) -> Bool {
+        SwiftFlutterBackgroundServicePlugin.taskIdentifier = "your.custom.task.identifier"
```
SputNikPlop, line 13:
```suggestion
        SwiftFlutterBackgroundServicePlugin.taskIdentifier = "com.rtirl.chat.tts"
```
@kevmo314 Is this what I should change it to or should it be left as default?

jlw4049 (approved):

kevmo314 (approved):

PR #1155, opened 2024-01-02
fix: update key and lints

SputNikPlop:


lib/components/chat_history/decorated_event.dart
```diff
@@ -13,14 +13,13 @@ class DecoratedEventWidget extends StatelessWidget {
   final EdgeInsets padding;
 
   const DecoratedEventWidget._(
-      {Key? key,
+      {super.key,
```
kevmo314, line 16: Is this what `dart fmt` spits out? It's fine if so, but a little surprised.

lib/components/chat_history/twitch/prediction_event.dart
```diff
@@ -36,11 +36,9 @@ class TwitchPredictionEventWidget extends StatelessWidget {
 
 class _TwitchOutcomeWidget extends StatelessWidget {
   const _TwitchOutcomeWidget(
-      {Key? key,
-      required this.outcome,
+      {required this.outcome,
```
kevmo314, line 39: Missing `super.key`

lib/components/header_bar.dart
```diff
@@ -11,8 +11,7 @@ class _DurationWidget extends StatelessWidget {
   final DateTime from;
   final TextStyle? style;
 
-  const _DurationWidget({Key? key, this.style, required this.from})
-      : super(key: key);
+  const _DurationWidget({this.style, required this.from});
```
kevmo314, line 14: Missing `super.key`.

kevmo314 (approved):

lib/components/chat_history/decorated_event.dart
SputNikPlop: Yeah dart format nor the analyzer throw me any errors at all, if I do try and correct it to the previous way the latest analyzer warns me to change it again.

lib/components/chat_history/twitch/prediction_event.dart
SputNikPlop: A value for optional parameter 'key' isn't ever given.
Try removing the unused parameter.

lib/components/header_bar.dart
SputNikPlop: A value for optional parameter 'key' isn't ever given.
Try removing the unused parameter.

PR #1158, opened 2024-01-04
fix: app bar icon colors

### LINKED ISSUE #1156: App bar color pallet is incorrectly set
This is in dark mode
Screen Shot 2024-01-03 at 1 53 58 PM

This in light mode
Screen Shot 2024-01-03 at 2 48 04 PM

We should have the app icons be white in both modes, this also applies to the qr icon on the sidebar.

SputNikPlop:
This should fix #1156 it needs to be tested, also let me know if there is a better implementation.


jlw4049 (approved):
Looks good!

kevmo314:
Re: simpler implementation, you could try setting the theme data at the AppBar: https://stackoverflow.com/a/57738275/86433

PR #1159, opened 2024-01-04
fix: update tts button

SputNikPlop:
This should take care of #1154 please double check it on your android simulator @jlw4049 when you have a chance.

### DIFF
--- a/lib/screens/home.dart
+++ b/lib/screens/home.dart
@@ -5,6 +5,7 @@ import 'package:flutter/material.dart';
 // import 'package:flutter_background_service/flutter_background_service.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:provider/provider.dart';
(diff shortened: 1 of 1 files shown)


lib/screens/home.dart
```diff
@@ -247,6 +248,12 @@ class _HomeScreenState extends State<HomeScreen> with TickerProviderStateMixin {
                           await TextToSpeechPlugin.clear();
                           // FlutterBackgroundService().invoke("stopService");
                         }
+
+                        final streamPrefs =
+                            await StreamingSharedPreferences.instance;
+                        streamPrefs.setString(
```
kevmo314, line 254: This should also be cleared when tts is disabled

kevmo314, line 254: This should also be cleared when tts is disabled.

kevmo314 (approved):

kevmo314 (approved):

PR #1162, opened 2024-01-04
fix: add vocalization for each message

### LINKED ISSUE #1152: For each message, add its vocalization to the speak system

SputNikPlop:
This should resolve #1152 though needs to be tested.


lib/components/chat_history/message.dart
```diff
@@ -75,11 +75,14 @@ class ChatHistoryMessage extends StatelessWidget {
         final loginChannelId = userModel.userChannel?.channelId;
         final viewingChannelId = m.channelId.split(':')[1];
         final channel = Channel("twitch", viewingChannelId, "");
+        final ttsModel = Provider.of<TtsModel>(context, listen: false);
 
         if (loginChannelId != viewingChannelId) {
           return child;
         }
 
+        ttsModel.say(message);
```
kevmo314, line 84: This shouldn't be here, the widget build function should be side effect free.

kevmo314 (changes requested):

jlw4049:
I think this is what @kevmo314 was referring to @SputNikPlop. 

Can you take a look at the changes I just added to #1162 and let me know if it looks like this is going down the correct path?

lib/tts_isolate.dart
```diff
@@ -62,25 +63,44 @@ void onStart(ServiceInstance service) async {
...
 
-Future<List<dynamic>> fetchMessagesFromFirestore(String channelId) async {
-  List<dynamic> messages = [];
+void listenForNewMessages(String channelId) {
+  // Listen to changes in Firestore and process messages
+  FirebaseFirestore.instance
+      .collection('channels')
+      .where('channelId', isEqualTo: channelId)
+      .snapshots()
+      .listen((QuerySnapshot<Map<String, dynamic>> snapshot) {
```
kevmo314, line 78: This is almost correct, nice work. The code as written now won't unsubscribe when the channel changes though. You can fix this with `.switchMap()` from the `stream_transforms` package in the caller:

```
import 'package:stream_transforms/stream_transforms.dart';

prefs.getString('tts_channel', default_value: '{}')
  .switchMap((channel) async {
    return FirebaseFirestore.instance
      .collection('channels')
      .where('channelId', isEqualTo: channel.id)
      .snapshots();
    })
  .listen((snapshot) {
    ...
  });
```

Here is the documentation for `switchMap()`: https://pub.dev/documentation/stream_transform/latest/stream_transform/Switch/switchMap.html

And here's a more insightful explanation: https://stackoverflow.com/questions/61569917/how-do-i-nest-streams-in-dart-map-streams-to-stream-events

kevmo314 (changes requested):

jlw4049:
@kevmo314 can you take a look at the latest changes and let me know how things are looking?

lib/tts_isolate.dart
```diff
@@ -59,28 +61,59 @@ void onStart(ServiceInstance service) async {
...
+          vocalizeMessage(message);
+        }
       }
     });
   });
 }
 
-Future<List<dynamic>> fetchMessagesFromFirestore(String channelId) async {
-  List<dynamic> messages = [];
+void listenForNewMessages(String channelId) {
```
kevmo314, line 85: This can be removed now, right?

lib/tts_isolate.dart
```diff
@@ -59,28 +61,59 @@ void onStart(ServiceInstance service) async {
 
   service.on('initSharedPreference').listen((event) async {
```
kevmo314, line 62: Nice work, we should also switchMap on this one too. Let me know if you'd like an example.

kevmo314 (approved):
Looks good just two minor comments.

lib/tts_isolate.dart
jlw4049: Yes haha, whoops!

lib/tts_isolate.dart
jlw4049: Yeah I might need an example for this one!

jlw4049: I _might_ have gotten it. Let me know if the last commit does the trick?

lib/tts_isolate.dart
```diff
@@ -59,28 +61,44 @@ void onStart(ServiceInstance service) async {
...
+          if (channel.isNotEmpty && channel != "{}") {
+            return FirebaseFirestore.instance
+                .collection('channels')
+                .where('channelId', isEqualTo: channel)
+                .snapshots();
+          } else {
+            return Stream.empty();
+          }
+        })
+        .switchMap((snapshot) => Stream.fromIterable(snapshot.docChanges))
```
kevmo314, line 77: Ah ok, so I was a little confused. It's not this change, but rather `service.on('initSharedPreferences')` isn't correct.

Actually, `flutter_background_service` already does the isolate handling for you, so all that needs to be done is instead of calling `service.on('...'), at the top level just add the listener. That is,

```
await Firebase.initializeApp();

final prefs = await ...;

prefs.getString(...).switchMap(...)...
```

I'd strongly recommend actually testing this by clicking the button and getting the tts engine to spit out text. This stuff is not easy to figure out without trying it.

jlw4049: Just pushed some changes, @SputNikPlop is going to test!

lib/tts_isolate.dart
```diff
@@ -43,10 +45,30 @@ Future<bool> onIosBackground(ServiceInstance service) async {
...
+          return Stream.empty();
+        }
+      })
+      .switchMap((snapshot) => Stream.fromIterable(snapshot.docChanges))
+      .listen((change) {
+        if (change.type == DocumentChangeType.added) {
+          var message = change.doc.data();
+          vocalizeMessage(message);
+        }
+      });
```
kevmo314, line 69:
```suggestion
      .listen((change) {
        for (final change in snapshot.docChanges) {
          if (change.type == DocumentChangeType.added) {
            vocalizeMessage(change.doc.data());
          }
        }
      });
```

There's no need to convert to a stream here.

kevmo314 (approved):
This looks good. For testing, try adding a button that starts the service and set the tts channel. You should hear it vocalize messages out of that channel.

jlw4049 (commented):
Tested it out, seems to be working properly

PR #1164, opened 2024-01-05
fix: better icon implementation

SputNikPlop:
@kevmo314 Is this what you wanted? If it is then I'll throw up another commit on this branch to remove the code from the other PR I added


kevmo314 (approved):
Yeah that's it. You can remove the individual color attributes I think with this.

SputNikPlop:
> Yeah that's it. You can remove the individual color attributes I think with this.

Done.

PR #1175, opened 2024-01-07
fix: stream preview volume

kevmo314:
The new twitch player does not appear to use an embedded iframe so we can control the video element directly.


SputNikPlop:
I wasn't able to get this to work and play audio like the function works on iOS

Issue #81, opened 2021-07-29
[chat_history] receive sharing intents


kevmo314:
It would be cool to receive sharing intents from other apps and populate the text field automatically. This plugin scaffolds the required steps: https://pub.dev/packages/receive_sharing_intent

It seems easier to not use the plugin though and instead write our own native code. We only need to handle text intents.

PR #1178, opened 2024-01-10
feat: receive sharing intents android

Wolfof420Street:
- receive sharing intents from other apps and populate the text field automatically.


SputNikPlop:
I went to wikipedia highlighted a noun and try to share it to the app it both copies the text as well as the url. I also tried sharing a text message to the chat app and it works as intended. Unless it is just quirks as android it partially works as intended.

SputNikPlop:
Ok this seems to work and the behavior is similar to other system apps. This is the Android part to #81

SputNikPlop (approved):
LGTM

PR #1179, opened 2024-01-15
fix: overflow errors

SputNikPlop:
This should fix the message input overflowing and the box I think being large.


kevmo314 (changes requested):
Screenshot?

SputNikPlop:
@kevmo314 I will put a video of the new behavior in the discord

kevmo314 (approved):

PR #1191, opened 2024-01-18
fix: add flash to scanner and improve ui

SputNikPlop:
This was it before
IMG_2188 Medium

and now after
IMG_2187 Medium


lib/screens/settings/audio_sources.dart
```diff
@@ -141,42 +143,122 @@ class _AudioSourcesScreenState extends State<AudioSourcesScreen> {
...
+                                                        IconButton(
+                                                            onPressed: () =>
+                                                                _scanController
+                                                                    .toggleTorch(),
+                                                            icon: icon),
+                                                      ],
+                                                    );
+                                                  },
+                                                ),
+                                              ),
```
kevmo314, line 248: Move this to a separate widget so it can be shared between the two scanner implementations.

SputNikPlop: Ok this makes more sense now.

lib/screens/settings/activity_feed.dart
```diff
@@ -151,42 +152,112 @@ class _ActivityFeedScreenState extends State<ActivityFeedScreen> {
...
-                                    },
+                                          Navigator.pop(ctx);
+                                        },
+                                      ),
+                                      Positioned(
+                                        top: 50,
+                                        left: 0,
+                                        right: 0,
+                                        child:
+                                            ValueListenableBuilder<TorchState>(
```
kevmo314, line 202: Replace these too

kevmo314 (changes requested):

SputNikPlop:
@kevmo314 Do I put the scanner_settings as a component or in the model? I don't think I should keep it under screens

lib/screens/settings/activity_feed.dart
SputNikPlop: Done

lib/screens/settings/scanner_settings.dart
```diff
@@ -0,0 +1,44 @@
+import 'package:flutter/material.dart';
+import 'package:mobile_scanner/mobile_scanner.dart';
+
+Widget scannerSettings(
```
kevmo314, line 4: This should be a full widget, not just a function that returns one.

```
class ScannerSettings extends StatelessWidget {
  @override
    ...
  }
}
```

lib/components/scanner_settings.dart
```diff
@@ -0,0 +1,50 @@
...
+              color: iconColor,
+            );
+            break;
+          case TorchState.off:
+            icon = const Icon(
+              Icons.flash_on,
+              color: iconColor,
+            );
+            break;
+        }
```
kevmo314, line 30: Inline this into the `const Icon()` below:

```
icon: Icon(
  value == TorchState.on ? Icons.flash_off : Icons.flash_on,
  color: iconColor,
),
```

kevmo314 (approved):

PR #1192, opened 2024-01-18
fix: theme data/text not showing in light mode

SputNikPlop:
This should fix the settings not being visible lemme know @kevmo314 if this is what you quite wanted.


lib/themes.dart
kevmo314, line 1: Move the other files to a separate change, this one is the only one that is needed to fix the issue.

SputNikPlop: Sorry about that. FIxed!

kevmo314 (approved):

PR #1193, opened 2024-01-19
fix: disable tts when stream goes offline

SputNikPlop:


lib/models/tts.dart
```diff
@@ -277,6 +277,13 @@ class TtsModel extends ChangeNotifier {
     if (!enabled && !force) {
       return;
     }
+
+    if (model is StreamStateEventModel) {
```
kevmo314, line 281: This should be done in the TTS isolate instead since the screen may not be open.

kevmo314 (changes requested):

lib/models/tts.dart
SputNikPlop: Good catch, thought it could be a general tts function.

lib/tts_isolate.dart
```diff
@@ -83,6 +83,16 @@ void vocalizeMessage(Map<String, dynamic>? message) {
     return;
   }
 
+  //check if message model is StreamStateEventModel
+  if (message.containsKey('isOnline')) {
+    bool isOnline = message['isOnline'] as bool;
+
+    if (!isOnline) {
+      TextToSpeechPlugin.stopSpeaking();
```
kevmo314, line 91: This is on the correct thread now, however it won't disable the icon either. The icon will still show that TTS is enabled. Think about what manages the state of TTS being enabled/disabled? This isolate listens on that state, we need to adjust the button to also listen on that state.

kevmo314 (changes requested):

jlw4049 (approved):
Looks good, I tested it on both physical/virtual android it is behaving correctly, although not sure how to simulate a stream going offline.

Edit: Doesn't disable notification or tts when stream dies.

lib/screens/home.dart
```diff
@@ -178,6 +181,17 @@ class _HomeScreenState extends State<HomeScreen> with TickerProviderStateMixin {
         model.showAudioPermissionDialog(context);
       }
     });
+    initSharedPreference();
+  }
+
+  initSharedPreference() async {
+    streamPrefs = await StreamingSharedPreferences.instance;
+    streamPrefs.getString('tts_channel', defaultValue: '{}').listen((value) {
```
kevmo314, line 189: Add an unsubscribe call in `dispose()`

lib/screens/home.dart
```diff
@@ -178,6 +181,17 @@ class _HomeScreenState extends State<HomeScreen> with TickerProviderStateMixin {
...
+      setState(() {
+        ttsChannel = value;
+        isTTsActive = value.isEmpty || value == "{}" ? false : true;
```
kevmo314, line 192: Make this a computed property instead

lib/screens/home.dart
```diff
@@ -237,42 +251,49 @@ class _HomeScreenState extends State<HomeScreen> with TickerProviderStateMixin {
...
+                            ? Icons.record_voice_over
+                            : Icons.voice_over_off),
+                        tooltip: AppLocalizations.of(context)!.textToSpeech,
+                        onPressed: () async {
+                          // Toggle the enabled state
 
-                        final streamPrefs =
-                            await StreamingSharedPreferences.instance;
+                          if (!isTTsActive) {
+                            ttsModel.enabled = true;
```
kevmo314, line 262: I don't think this line does anything, does it work with it removed?

lib/screens/home.dart
```diff
@@ -237,42 +251,49 @@ class _HomeScreenState extends State<HomeScreen> with TickerProviderStateMixin {
...
-                              body: null,
-                              locked: true,
-                              autoDismissible: false,
-                            ),
-                          );
-                        }
-                      },
-                    );
+                          if (isTTsActive) {
+                            ttsModel.enabled = false;
```
kevmo314, line 285: Same here

kevmo314 (changes requested):
Nice, logic looks good just a few cleanup points.

lib/tts_isolate.dart
```diff
@@ -36,9 +36,21 @@ Future<void> isolateMain(
           .snapshots()
           .listen((latestMessage) async {
         if (latestMessage.docs.isNotEmpty) {
-          final textToSpeak = latestMessage.docs[0]['message'] as String?;
-          if (textToSpeak != null) {
-            await ttsQueue.speak(latestMessage.docs[0].id, textToSpeak);
+          final message = latestMessage.docs[0];
+          if (message.data().containsKey('type')) {
+            final type = message['type'] as String?;
+            if (type == "message") {
```
kevmo314, line 42: Use a switch here

```
switch (type) {
  case "message": ...
  case "stream.offline": ...
}
```

kevmo314 (approved):
Nice work, make sure the notifications change gets looped in with @SputNikPlop's work, otherwise this looks great.

PR #1194, opened 2024-01-19
feat: add button to disable tts

SputNikPlop:
This should add a button to disable tts. I added an icon instead of the text because I was unsure which to add so lemme know if I need to go to text or if you want to switch the icon to something like https://fonts.google.com/icons?selected=Material%20Symbols%20Outlined%3Avoice_over_off%3AFILL%400%3Bwght%40400%3BGRAD%400%3Bopsz%4024


lib/tts_isolate.dart
```diff
@@ -47,6 +73,58 @@ void onStart(ServiceInstance service) async {
...
+      service.setAsBackgroundService();
+    });
+  }
+
+  service.on('stopService').listen((event) {
+    service.stopSelf();
+  });
+
+  //bring to foreground
+  Timer.periodic(const Duration(seconds: 1), (timer) async {
```
kevmo314, line 103: This will burn a lot of battery, I think the call only needs to be made once? Did you take this code from somewhere else?

kevmo314 (changes requested):

lib/tts_isolate.dart
SputNikPlop: I was playing around with others ways to handle this, why I thought this way would be better for android but if the cost is too high lemme try and change it to my other way.

android/app/src/main/res/drawable/voiceover.png
SputNikPlop, line 1: @kevmo314 Is this the right image you want?

lib/tts_isolate.dart
```diff
@@ -68,15 +120,46 @@ void onStart(ServiceInstance service) async {
...
+    TextToSpeechPlugin.stopSpeaking();
     service.stopSelf();
   });
 
   service.on('startTts').listen((event) async {
     // print("Starting this service right now");
     await Isolate.spawn(isolateMain, ReceivePort().sendPort);
+
+    //bring service to foreground
+    bringServiceToForeground(service, flutterLocalNotificationsPlugin);
```
kevmo314, line 132: Does this work? This is being called from the wrong isolate.

android/app/src/main/kotlin/com/rtirl/chat/NotificationService.kt
```diff
@@ -0,0 +1,124 @@
...
+                        .addAction(R.drawable.text_to_speech, "Disable TTS", disablePendingIntent)
+                .build()
+
+                startForeground(NOTIFICATION_ID, notification)
+
+                Log.d("NotificationService", "startForeground called")
+
+
+                // Do heavy work on a background thread
+                // stopSelf();
```
kevmo314, line 80:
```suggestion
```

android/app/src/main/kotlin/com/rtirl/chat/NotificationService.kt
```diff
@@ -0,0 +1,124 @@
...
+import android.content.Context
+import android.content.Intent
+import android.os.Build
+import android.os.IBinder
+import android.speech.tts.TextToSpeech
+import android.util.Log
+import androidx.core.app.NotificationCompat
+import com.rtirl.chat.R
+
+class NotificationService : Service() {
```
kevmo314, line 16: Does the flutter notification lib we use support a button? I'm surprised this has to be done manually.

SputNikPlop: Not quite sure how it works on android and I might have messed this up but this was my attempt to ditch awesome_notifications

jlw4049 (approved):

android/app/src/main/kotlin/com/rtirl/chat/TextToSpeechSingleton.kt
```diff
@@ -0,0 +1,25 @@
+package com.rtirl.chat
```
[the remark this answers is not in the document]
SputNikPlop: I think so, I'm submitting a fix for that

PR #1197, opened 2024-01-20
fix: add raid visibility

SputNikPlop:
This adds a raid button on the sidebar to further use and help people find the raid button.

### DIFF
--- a/lib/components/drawer/sidebar.dart
+++ b/lib/components/drawer/sidebar.dart
@@ -212,6 +212,48 @@ class _SidebarState extends State<Sidebar> {
 
       const Divider(),
 
+      // setting
+      Consumer<LayoutModel>(builder: (context, layoutModel, child) {
+        if (!layoutModel.locked) {
+          return ListTile(
+            leading: const Icon(Icons.thunderstorm),
+            title: Text(AppLocalizations.of(context)!.enableRainMode),
+            subtitle:
+                Text(AppLocalizations.of(context)!.enableRainModeSubtitle),
+            onTap: () async {
+              layoutModel.locked = !layoutModel.locked;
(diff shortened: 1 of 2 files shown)


lib/components/drawer/sidebar.dart
```diff
@@ -212,6 +212,56 @@ class _SidebarState extends State<Sidebar> {
 
       const Divider(),
```
SputNikPlop, line 213: Not sure if you wanted a divider or where to put so asked in the discord.

lib/components/drawer/sidebar.dart
```diff
@@ -212,6 +212,56 @@ class _SidebarState extends State<Sidebar> {
...
 
+      //raid
+      Consumer<UserModel>(builder: (context, model, child) {
+        return ListTile(
+          leading: const Icon(Icons.connect_without_contact),
+          title: Text(AppLocalizations.of(context)!.raidAChannel),
```
SputNikPlop, line 219: Sidebar only says Raid instead of Raid Channel because then I would have to add another localization if you want me to change that lemme know.

lib/components/drawer/sidebar.dart
```diff
@@ -212,6 +212,56 @@ class _SidebarState extends State<Sidebar> {
 
       const Divider(),
 
+      //raid
+      Consumer<UserModel>(builder: (context, model, child) {
+        return ListTile(
+          leading: const Icon(Icons.connect_without_contact),
```
SputNikPlop, line 218: Let me know if you want a better icon and which I should choose.

SputNikPlop (commented):
@kevmo314 Things that I think should be reviewed.

PR #1202, opened 2024-01-26
fix: disable tts on low battery

SputNikPlop:
The point of this pr is to disable the tts when the battery drops below five percent unless charging.


SputNikPlop:
The button should also be disabled for tts to turn it on when the battery level is too low

SputNikPlop:
As mentioned https://github.com/muxable/rtchat/pull/1204#discussion_r1723009029 not worth increasing bundle size

PR #1204, opened 2024-01-26
fix: stream preview power

SputNikPlop:
The intent of this is if the battery falls below 20% not charging and not in battery saving mode show the prompt that stream preview takes up lots of battery, then if the battery falls below 5% disable the stream preview if open


SputNikPlop:
@kevmo314 Anything else to add to this?

pubspec.yaml
```diff
@@ -11,6 +11,7 @@ dependencies:
   flutter:
     sdk: flutter
   audioplayers: ^6.0.0
+  battery_plus: ^5.0.2
```
kevmo314, line 14: Hmm I think the idea is interesting but I don't think it's worth further increasing our bundle size for this. Right now the prompt shows after a timer which wouldn't have any dependency impact.

SputNikPlop: This though would be shared anyway if #1202 would still be implemented. Same thoughts on that one as well?

PR #1210, opened 2024-02-09
fix: connect tts queue

jlw4049:


lib/tts_isolate.dart
```diff
@@ -104,35 +93,74 @@ void onStart(ServiceInstance service) async {
...
-    try {
-      TextToSpeechPlugin.speak(textToSpeak);
-    } catch (e) {
-      debugPrint("Error during TTS processing: $e");
-    }
-  } else {
-    debugPrint("Received a message with null or missing 'text' field");
-  }
-}
+// void vocalizeMessage(Map<String, dynamic>? message) async {
```
kevmo314, line 141: Remove commented code

lib/tts_isolate.dart
```diff
@@ -52,44 +56,29 @@ void onStart(ServiceInstance service) async {
   DartPluginRegistrant.ensureInitialized();
   WidgetsFlutterBinding.ensureInitialized();
   await Firebase.initializeApp();
+  // final prefs = await StreamingSharedPreferences.instance;
```
kevmo314, line 59: We should listen on the prefs instead of the tts channel passed through the argument. So in order to start tts, we should do two things:

1. Call `FlutterBackgroundService().invoke('startTts');` like you're doing now, but don't pass the channel.
2. Update shared prefs `tts_channel` to the tts channel json.

This is so we can update the channel regardless of if the service is started or not. Otherwise, right now you can't switch channels with tts on (well, you can but tts won't carry over).

kevmo314 (changes requested):
Nice work figuring this out. Remove the commented out code, otherwise looks good just one detail on how the channel is passed.

lib/tts_isolate.dart
jlw4049: I had one issue attempting to persist data from `prefs` from main isolate to tts isolate. Initializing the TTS would actually update the value I had tested with from false to true, but disabling it, no matter how many times I tried, the data would stay `true` and never kill the listen service. 

I have a lot less knowledge on this than you, but to my understanding each isolate has it's own cache. I am assuming there is a way to work around this, maybe with a singleton class or something? What do you suggest? 

Also would it be better to stop using `streaming_shared_preferences` and use `shared_preferences` instead as it looks like the former was abandoned some years ago. 

Here is what I've [found](https://stackoverflow.com/a/75258254/13284500) on the subject, let me know your thoughts before I continue here.

kevmo314: Interesting, well that fuckin' sucks. Ok, so then let's try to use the `invoke` in a slightly different way: instead of `invoke("startTts", ...)` let's make it just `invoke("setTtsChannel", ...);` and if you want to stop TTS, do `invoke("setTtsChannel", { channel: null });`.

In the isolate, we should listen on `setTtsChannel` and if we receive a null, stop listening and exit the thread. This would look something like:

```
Subscription? messagesSubscription = null;
Subscription? channelSubscription = bgchannel.on("setTtsChannel").listen((event) => {
  if (event.channel == null) {
    channelSubscription.unsubscribe();
  }.else {
    messagesSubscription?.unsubscribe();
    messagesSubscription = Firestore.collection(...).snapshots().listen((data) => {
      // read message
    });
  }
});
```

This should exit the isolate when `channel: null` is sent.

kevmo314: Re: removing `streaming_shared_preferences` yeah that sounds reasonable if we do the above approach. I had hoped `streaming_shared_preferences` would work as a way to communicate between the isolates but I guess not. :(

jlw4049: If you can take a look at the latest commit I did based on your response to see if it's what you'd like. Let me know if there needs to be any refactoring. I tested it, it's working as well as the other way I had it. 

A couple things I was unsure of, if it was safe to put 
```
StreamSubscription? messagesSubscription;
StreamSubscription? channelSubscription;
```
at the top of the file like I did, or if it was okay to be be in the function. 

Also you had it this way, 
```dart
  if (event.channel == null) {
    channelSubscription.unsubscribe();
  }.else {
  ```
  
  To make it work it had to be like below. If I canceled the `channelSubscription` it would not allow it to work again when re enabling the TTS.
  ```dart
  if (args['channel'] == null) {
        await ttsQueue.clear();
        messagesSubscription?.cancel();
      } else {
      ```
Just wanted to make sure that was alright?

Comment (jlw4049) [reply lib/tts_isolate.dart]: Also completely unrelated, this doesn't happen outside of the TTS, but might be worth looking into or it might be safe to ignore. But when you disable TTS (regardless of the isolate code) you get this warning. 

image

File: lib/tts_isolate.dart [hunk 3]
```diff
@@ -53,41 +54,35 @@ void onStart(ServiceInstance service) async {
...
+            .doc(ttsCurrentChannel)
+            .collection('messages')
+            .orderBy('timestamp', descending: true)
+            .limit(1)
+            .snapshots()
+            .listen((latestMessage) async {
+          if (latestMessage.docs.isNotEmpty) {
+            final timestamp = latestMessage.docs[0]['timestamp'];
+            final userId = latestMessage.docs[0]['author']["userId"];
+            final uniqueId = '$timestamp-$userId';
```
kevmo314, line 79: You can use `doc.id`

lib/tts_isolate.dart
```diff
@@ -53,41 +54,35 @@ void onStart(ServiceInstance service) async {
...
-      if (change.type == DocumentChangeType.added) {
-        vocalizeMessage(change.doc.data());
+  channelSubscription = service.on("setTtsChannel").listen((event) async {
+    if (event != null) {
+      Map<String, dynamic>? args = event;
+      if (args['channel'] == null) {
+        await ttsQueue.clear();
+        messagesSubscription?.cancel();
+      } else {
+        var channelData = args["channel"];
```
kevmo314, line 64: final

lib/tts_isolate.dart
```diff
@@ -53,41 +54,35 @@ void onStart(ServiceInstance service) async {
...
+            "${channelData['provider']}:${channelData['channelId']}";
+        messagesSubscription?.cancel();
+        messagesSubscription = FirebaseFirestore.instance
+            .collection('channels')
+            .doc(ttsCurrentChannel)
+            .collection('messages')
+            .orderBy('timestamp', descending: true)
+            .limit(1)
+            .snapshots()
+            .listen((latestMessage) async {
```
kevmo314, line 75: I think there's a subtle bug here that this will read the last message whereas we want to only read future messages. The way around this is to keep a timestamp when the listener is added and reject messages that are added before that timestamp.

kevmo314 (approved):
Nice work looks great!

lib/tts_isolate.dart
jlw4049: ah nice!

lib/tts_isolate.dart
jlw4049: This should resolve it

kevmo314 (approved):

SputNikPlop (approved):
This should be good to go

PR #1225, opened 2024-03-21
fix: update get vocalization in the isolate

### LINKED ISSUE #1188: Update getVocalization() to use this function
https://github.com/muxable/rtchat/blob/2ca0e7224ff603c3a8e210fa20a8ddda01cafcae/lib/models/tts.dart#L108

We should call this function instead. For this change, we'll want to instantiate a `TtsModel` when the isolate is created and call the function on that model.

jlw4049:
Fixes #1188

### DIFF
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -73,8 +73,9 @@ void main() async {
   await Firebase.initializeApp(options: DefaultFirebaseOptions.currentPlatform);
   await MobileAds.instance.initialize();
   final prefs = await StreamingSharedPreferences.instance;
+  final ttsModel = TtsModel.emptyConstructor();
   await tts_isolate.isolateMain(
-      ReceivePort().sendPort, channelStreamController, prefs);
+      ReceivePort().sendPort, channelStreamController, prefs, ttsModel);
 
   if (!kDebugMode) {
     FlutterError.onError = FirebaseCrashlytics.instance.recordFlutterFatalError;
--- a/lib/main.dart
+++ b/lib/main.dart
@@ -73,9 +73,8 @@ void main() async {
   await Firebase.initializeApp(options: DefaultFirebaseOptions.currentPlatform);
   await MobileAds.instance.initialize();
   final prefs = await StreamingSharedPreferences.instance;
-  final ttsModel = TtsModel.emptyConstructor();
   await tts_isolate.isolateMain(
-      ReceivePort().sendPort, channelStreamController, prefs, ttsModel);
+      ReceivePort().sendPort, channelStreamController, prefs);
 
   if (!kDebugMode) {
     FlutterError.onError = FirebaseCrashlytics.instance.recordFlutterFatalError;
--- a/lib/tts_isolate.dart
+++ b/lib/tts_isolate.dart
@@ -83,7 +83,9 @@ Future<void> isolateMain(
                   messageModel,
                   includeAuthorPrelude: !isPreludeMuted,
                 );
-                await ttsQueue.speak(message.id, finalMessage);
+                if (finalMessage.isNotEmpty) {
+                  await ttsQueue.speak(message.id, finalMessage);
+                }
                 break;
               case "stream.offline":
                 await ttsQueue.clear();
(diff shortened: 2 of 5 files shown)


lib/main.dart
```diff
@@ -73,8 +73,9 @@ void main() async {
   await Firebase.initializeApp(options: DefaultFirebaseOptions.currentPlatform);
   await MobileAds.instance.initialize();
   final prefs = await StreamingSharedPreferences.instance;
+  final ttsModel = TtsModel.emptyConstructor();
```
kevmo314, line 76: Does this actually work? The `TtsModel` should be initialized with `fromJson()` read from shared preferences.

lib/tts_isolate.dart
```diff
@@ -24,14 +27,50 @@ Future<void> isolateMain(
...
-  // listen for changes to the tts preferences
-  bool isPreludeMuted = false;
+  final ttsModel = TtsModel.fromJson(
+      jsonDecode(prefs.getString("tts", defaultValue: '{}').getValue()));
+
+  // listen for changes to the tts preferences and update the isolates ttsModel
   final ttsPrefs = prefs.getString('tts', defaultValue: '{}');
-  ttsPrefs.listen((value) {
+  ttsPrefs.listen((value) async {
     final Map<String, dynamic> ttsMap = jsonDecode(value);
```
kevmo314, line 36: Does something like

```
final ttsModel = TtsModel.fromJson(jsonDecode(value));
```

work here? That would allow avoiding adding all the setters.

Also I realized this suffers from the same shared preferences caching issue you mentioned before, you may have to route the preferences through the method channel instead of listening on shared preferences.

lib/tts_isolate.dart
```diff
@@ -48,28 +76,33 @@ Future<void> isolateMain(
...
-                  final userName = message['author']['displayName'] as String?;
-                  finalMessage = isPreludeMuted
-                      ? textToSpeak
-                      : (userName != null
-                          ? '$userName said $textToSpeak'
-                          : textToSpeak);
+                final timestamp =
+                    (messageData['timestamp'] as Timestamp?)?.toDate() ??
+                        DateTime.now();
+                final messageModel = createMessageModelFromData(
```
kevmo314, line 90: Inline the function

lib/tts_isolate.dart
```diff
@@ -24,14 +27,39 @@ Future<void> isolateMain(
...
-    isPreludeMuted = ttsMap['isPreludeMuted'];
+  ttsPrefs.listen((value) async {
+    ttsModel.updateFromJson(jsonDecode(value));
+    await TextToSpeechPlugin.updateTTSPreferences(
+        ttsModel.pitch, ttsModel.speed);
   });
 
+  TwitchMessageModel createMessageModelFromData(
+      String messageId, Map<String, dynamic> messageData, DateTime timestamp) {
+    return TwitchMessageModel(
```
kevmo314, line 43: This mapping is almost correct, you can simplify it a little bit. Steal the code from here: https://github.com/muxable/rtchat/blob/96b8798af1db8cd73c5197d927ece0eae951680e/lib/models/adapters/messages.dart#L97-L108

kevmo314 (approved):
Looks good!

SputNikPlop (approved):
thanks for taking care of the android bit

lib/tts_isolate.dart
```diff
@@ -48,28 +57,56 @@ Future<void> isolateMain(
...
+                            ),
+                          )
+                        : null,
+                    tags: messageData['tags'],
+                    annotations: TwitchMessageAnnotationsModel.fromMap(
+                        messageData['annotations']),
+                    thirdPartyEmotes: await getEmotes(Channel(
+                        messageData['channelId'].toString().split(":")[0],
+                        messageData['channelId'],
+                        messageData['author']['displayName'])),
```
kevmo314, line 92:
```suggestion
                    thirdPartyEmotes: {},
```

Ah, this is a bit unfortunate, we shouldn't fetch emotes like this repeatedly every channel as it will use too much data. The good news is that we don't really need third-party emotes for tts so you can wipe the map and pass an empty map.

kevmo314 (approved):

PR #1244, opened 2024-05-10
fix: enable option to test new tts

SputNikPlop:
The new text to speech wasn't quite a hit but this should offer an option for this to be changed via the text to speech settings.

### DIFF
--- a/lib/screens/settings/tts.dart
+++ b/lib/screens/settings/tts.dart
@@ -262,7 +262,14 @@ class TextToSpeechScreen extends StatelessWidget {
               onChanged: (value) {
                 model.isPreludeMuted = value;
               },
-            )
+            ),
+            SwitchListTile.adaptive(
+              title: const Text("Try new text to speech"),
+              value: model.useNewTts,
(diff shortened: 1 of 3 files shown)


lib/screens/settings/tts.dart
```diff
@@ -262,7 +262,14 @@ class TextToSpeechScreen extends StatelessWidget {
               onChanged: (value) {
                 model.isPreludeMuted = value;
               },
-            )
+            ),
+            SwitchListTile.adaptive(
```
kevmo314, line 266: I don't think we should offer this as an option, nobody will switch to it. We should improve tts based on the feedback and then that would be enough.

kevmo314 (changes requested):

lib/screens/settings/tts.dart
kevmo314: In particular it seems like most people complained about the speed control not working and tts reading bot messages.

SputNikPlop:
Closing based on @kevmo314 feedback

PR #1247, opened 2024-05-21
fix: landscape qr

SputNikPlop:
This is to fix the qr code not working horizontally, it isn't quite perfect I think but I'll let you take a look at the following pictures.

The old qr:
Portrait - I like how the text is above the bottom navigation bar here.
IMG_2915 Medium

Landscape - It overflow and the qr code doesn't work because the profile picture is way to big.
IMG_2916 Medium

The new qr:
Portrait - The text is now very low and behind the bottom navigation bar.
IMG_2913 Medium

Landscape - The qr now works but I think the text is a little big.
IMG_2914 Medium

I like how the old portrait text was and kind of how it looked in landscape but the qr code didn't work, this new way works but I don't quite like the text. Maybe we make the text smaller in the new landscape ui matching the old way without having to scroll and being over the navigation bar in portrait mode. I think I am missing something though, let me know the best way to do this.

### DIFF
--- a/lib/screens/settings/qr.dart
+++ b/lib/screens/settings/qr.dart
@@ -11,8 +11,11 @@ class QRDisplay extends StatelessWidget {
   Widget build(BuildContext context) {
     final querySize = MediaQuery.of(context).size;
 
+    final isLandScape = querySize.width > querySize.height;
+
     return Consumer<UserModel>(builder: (context, userModel, child) {
       final userChannel = userModel.userChannel;
+
       final inviteLink =
           "https://www.twitch.tv/${userModel.userChannel?.displayName ?? ""}";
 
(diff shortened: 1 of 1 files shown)


SputNikPlop:
@kevmo314 I think this is ready now here is how it looks:
IMG_2920 Medium

SputNikPlop:
Oh here it is in landscape
IMG_2921 Medium

kevmo314:
> Oh here it is in landscape IMG_2921 Medium

Can you make the width fit the content for landscape? Otherwise this looks good.

SputNikPlop:
> Can you make the width fit the content for landscape? Otherwise this looks good.

So just remove all the excess white got it.

SputNikPlop:
@kevmo314 Take a look
IMG_2940 Medium

kevmo314 (approved):

PR #1250, opened 2024-06-03
chore: bump audioplayers version

kevmo314:
Adds more explicit audio annotations so maybe it will behave better with Spotify and other applications.


SputNikPlop:
Maybe we should do this?
https://developer.android.com/media/optimize/audio-focus

SputNikPlop (approved):
This helps somewhat, I tried a more in depth fix on my pr since eventually we have to handle alerts and more

PR #1251, opened 2024-06-09
fix: background audio handling

SputNikPlop:
This should better handle the audio while tts is on, should be tested though.


kevmo314:
Superficially this seems reasonable but I think we should launch the new TTS engine before getting this in. This adds a good amount of complexity and it would be better to simplify TTS and get the new engine out before making it more complex.

SputNikPlop:
> Superficially this seems reasonable but I think we should launch the new TTS engine before getting this in. This adds a good amount of complexity and it would be better to simplify TTS and get the new engine out before making it more complex.

Got it, there also is a slight difference between https://github.com/muxable/rtchat/pull/1251/commits/19ba377a7176d4e3450591c38b3f8e2bd5a3a87c and https://github.com/muxable/rtchat/pull/1251/commits/c1444cd87071f406ff0256c262a9504a83e051cf on the audio it seems so later that will have to be sorted to which version is correct

PR #1255, opened 2024-06-17
chore: upgrade scanner

SputNikPlop:
This upgrades the scanner to match the changes that got done in rtirl

### DIFF
--- a/lib/components/scanner_settings.dart
+++ b/lib/components/scanner_settings.dart
@@ -8,12 +8,8 @@ class ScannerSettings extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
-    return ValueListenableBuilder<TorchState>(
-      valueListenable: scanController.torchState,
-      builder: (context, value, child) {
-        const Color iconColor = Colors.white;
-
-        return Row(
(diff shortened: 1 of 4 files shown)


lib/components/scanner_settings.dart
```diff
@@ -8,32 +8,28 @@ class ScannerSettings extends StatelessWidget {
...
-              onPressed: () => scanController.toggleTorch(),
-              icon: Icon(
-                value == TorchState.on ? Icons.flash_off : Icons.flash_on,
-                color: iconColor,
-              ),
-            ),
-          ],
-        );
-      },
+    const Color iconColor = Colors.white;
```
kevmo314, line 11: I think we still need to listen on `scanController` here, otherwise the widget tree won't pick up the new torch state.

kevmo314 (changes requested):

lib/components/scanner_settings.dart
SputNikPlop: Okay, seems to work now.

kevmo314 (approved):

PR #1260, opened 2024-06-19
ci: build app on pull request

kevmo314:


SputNikPlop:
Will this mean now when we get iOS builds working to main that the version will automatically get pushed up to TestFlight making field testing easier?

PR #1261, opened 2024-06-19
fix: upgrade podfile

kevmo314:


SputNikPlop:
Should we not roll out releases right now to the last version which is compatible with iOS 12 the flutter minimum or would you want me to put everything now as the iOS 13 minimum which flutter will soon migrate to?

kevmo314:
> Should we not roll out releases right now to the last version which is compatible with iOS 12 the flutter minimum or would you want me to put everything now as the iOS 13 minimum which flutter will soon migrate to?

Yeah, one of the dependencies (firebase) requires ios 13 so everything needs to move over.

PR #1287, opened 2024-07-09
fix: emote/keyboard overflow and picker

SputNikPlop:
This fixes and overflow when swapping between the picker and the keyboard. Also fixes when an emote is too large (will be changed later).

This also changes the emote picker to look more like the old one and twitch's. The old one in our app.
IMG_74AADF11CE23-1

My new implementation.
IMG_3271

Twitch's implementation.
IMG_3271

Lemme know what I can tweak to make it better since this isn't 100% complete


kevmo314 (approved):
Looks good, seems reasonable thanks!

PR #1293, opened 2024-07-30
Add readiness & liveness status check handlers

charmitro:
This PR adds readiness and liveness status check handlers to the agent, which help ensure the health of the application in 
 Kubernetes.

Changes Made:
* Added` /readiness` endpoint to check if the service is ready.
* Added `/liveness` endpoint to check if the service is running.
* Kept the existing root endpoint handler. (not sure if this is OK)

It should work in Kubernetes like,
```yaml
        readinessProbe:
          httpGet:
            path: /readiness
            port: 8080
          initialDelaySeconds: 5
          periodSeconds: 10
        livenessProbe:
          httpGet:
            path: /liveness
            port: 8080
          initialDelaySeconds: 5
          periodSeconds: 10
```

**Sadly I cannot test this because it needs some Google Cloud auth:**
```
2024-07-29T23:01:12.775+0300    FATAL   cmd/main.go:92  failed to create firestore client       {"error": "google: could not find default credentials. See https://cloud.google.com/docs/authentication/external/set-up-adc for more information"}
```

### DIFF
--- a/agent/cmd/main.go
+++ b/agent/cmd/main.go
@@ -127,6 +127,18 @@ func main() {
 		}
 		w.Write([]byte((72*time.Hour - time.Since(start)).String()))
 	}))
+	http.Handle("/readiness", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if !isListening.Load() || time.Since(start) > 72*time.Hour {
+			w.WriteHeader(http.StatusServiceUnavailable)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+		w.Write([]byte("OK"))
+	}))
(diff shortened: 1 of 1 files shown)


charmitro:
I believe CI fails because it's a fork of the repository?

kevmo314:
Thanks for the PR! The ticket could've been more specific but the additional endpoints aren't actually necessary. We also don't use Kubernetes to deploy, really we only need some configuration set up in the GCP console. That being said, since the agent code hasn't changed much in the last few years, the utility of doing so it not very high.

I appreciate the help though, if you're interested we'd be more than happy to work in any frontend improvements you can think of.

PR #1298, opened 2024-08-03
feat: retry future up to three times with a one second delay

kevmo314:
Add retry logic to the `getTwitchBadges` method in `lib/models/adapters/chat_state.dart` to handle the `firebase_functions/unavailable` error.

* Use a loop to attempt the Firebase function call up to three times.
* Add a one second delay between each retry attempt.
* Catch the `firebase_functions/unavailable` error and retry if the attempt count is less than three.
* Return an empty list if all attempts fail.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/muxable/rtchat?shareId=99fff9a3-f4b8-4546-90d2-efd7d2b9dc06).

### DIFF
--- a/lib/models/adapters/chat_state.dart
+++ b/lib/models/adapters/chat_state.dart
@@ -99,31 +99,42 @@ class ChatStateAdapter {
     });
   }
 
-  Future<List<TwitchBadgeInfo>> getTwitchBadges({String? channelId}) {
-    return functions
-        .httpsCallable("getBadges")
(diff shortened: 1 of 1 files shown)


kevmo314:
@SputNikPlop can you verify this works? Just check that badges still load.

SputNikPlop:
> @SputNikPlop can you verify this works? Just check that badges still load.

They all still load

SputNikPlop (approved):

PR #1294, opened 2024-07-31
fix: autofill emotes while typing

SputNikPlop:
Incomplete need another pair of eyes on this.

Screenshot 2024-07-31 at 9 58 05 AM
This is what I am attempting to do, currently the cursor jumps to the start when backspacing and not sure how to quite handle the text replacement.

PR #1309, opened 2024-08-11
feat: add emote rendering in message input

kevmo314:
Play around with this and lmk what you think.


SputNikPlop (approved):
This looks good, some minor feedback I had was to make it look like the way I tried in #1294 for the picker, as well holding on the selector popping up the text name of the emote like in the global selector.
IMG_3435 Medium
This though I changed in #1287 I think

PR #1319, opened 2024-08-21
fix: image quality

SputNikPlop:
Not sure if this would help or if you have any suggestions on improving this

### DIFF
--- a/lib/components/autocomplete.dart
+++ b/lib/components/autocomplete.dart
@@ -113,6 +113,7 @@ class _AutocompleteWidgetState extends State<AutocompleteWidget> {
                     icon: Image(
                         width: 24,
                         height: 24,
+                        filterQuality: FilterQuality.high,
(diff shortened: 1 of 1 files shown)


kevmo314 (commented):
It says medium is the best for downscaling? https://api.flutter.dev/flutter/dart-ui/FilterQuality.html

If the emotes don't look high quality we probably need to download a higher res

SputNikPlop:
> It says medium is the best for downscaling? https://api.flutter.dev/flutter/dart-ui/FilterQuality.html
> 
> If the emotes don't look high quality we probably need to download a higher res

That should have been medium, its not that the emotes don't look high res enough it was just finding the right balance since it felt like some animated ones slowed things down. I wasn't sure enough which is why I tried this tweak and thought I would get more input from you.

kevmo314 (approved):

PR #1322, opened 2024-09-06
fix: improve landscape keyboard behavior

kevmo314:


SputNikPlop (approved):
I put my feedback in the discord, the gist is autocomplete should have probably the same amount of emotes (10) and the slash commands have an overflow which the logic on how to handle that maybe is display them without the description.

PR #1329, opened 2025-01-20
fix: gradle causing builds to fail

SputNikPlop:
Upgraded one version to fix the build failing or I could do maybe the workaround fix mentioned on https://issuetracker.google.com/issues/242308990
```┌─ Flutter Fix ────────────────────────────────────────────────────────────────────────────────────┐
│ [!] Version 7.3 of the Android Gradle Plugin (AGP) uses a version of R8 that contains a bug      │
│ which causes this error (see more info at https://issuetracker.google.com/issues/242308990).     │
│ To fix this error, update to a newer version of AGP (at least 7.4.0).                            │
│                                                                                                  │
│  The version of AGP that your project uses is likely defined in:                                 │
│ /home/runner/work/rtchat/rtchat/android/settings.gradle,                                         │
│ in the 'plugins' closure (by the number following "com.android.application").                    │
│  Alternatively, if your project was created with an older version of the templates, it is likely │
│ in the buildscript.dependencies closure of the top-level build.gradle:                           │
│ /home/runner/work/rtchat/rtchat/android/build.gradle,                                            │
│ as the number following "com.android.tools.build:gradle:".                                       │
└──────────────────────────────────────────────────────────────────────────────────────────────────┘```

### DIFF
--- a/android/gradle/wrapper/gradle-wrapper.properties
+++ b/android/gradle/wrapper/gradle-wrapper.properties
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6.4-all.zip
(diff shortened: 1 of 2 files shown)


kevmo314 (commented):
This doesn't actually bump the version of gradle, the version of gradle we use is configured here: https://github.com/muxable/rtchat/blob/main/android/gradle/wrapper/gradle-wrapper.properties#L6

SputNikPlop:
> This doesn't actually bump the version of gradle, the version of gradle we use is configured here: https://github.com/muxable/rtchat/blob/main/android/gradle/wrapper/gradle-wrapper.properties#L6

Ok changed that and the buildscript

PR #1333, opened 2025-02-14
fix:fix android build issues

Wolfof420Street:
- fox android build issues
- replaced motion senso dependency


kevmo314:
It looks like there's quite a lot of files in here that aren't related to your change, could you remove them?

Wolfof420Street:
@kevmo314  removed thee unnecessary files

lib/urls.dart
```diff
@@ -3,9 +3,9 @@ import 'package:url_launcher/url_launcher.dart';
 
 Future<void> openUrl(Uri url) async {
   if (!url.hasScheme) {
-    await customtabs.launch(url.replace(scheme: 'http').toString());
+    await customtabs.launchUrl(url.replace(scheme: 'http'));
   } else if (url.isScheme("http") || url.isScheme("https")) {
-    await customtabs.launch(url.toString());
+    await customtabs.launchUrl(url);
```
SputNikPlop, line 8: This won't have the same function as the previous function, instead it will prompt the user which web browser they want to open the link in.

lib/components/chat_history/twitch/message.dart
```diff
@@ -153,22 +153,21 @@ class TwitchMessageWidget extends StatelessWidget {
   Widget authorWidget(
       TwitchUserModel author, StyleModel styleModel, TextStyle authorStyle) {
     if (contributors.contains(model.author.userId)) {
-      return StreamBuilder<AbsoluteOrientationEvent>(
-          stream: motionSensors.absoluteOrientation,
+      return StreamBuilder<AccelerometerEvent>(
```
SputNikPlop, line 156: This method is deprecated, also I think there was a difference in return value which is why sensors_plus wasn't used before

pubspec.yaml
```diff
@@ -19,44 +19,42 @@ dependencies:
...
-  just_audio: ^0.9.39
+  just_audio: ^0.9.46
   linkify: ^5.0.0
   metadata_fetch: ^0.4.1
-  mobile_scanner: ^5.0.2
-  motion_sensors:
-    git:
-      url: https://github.com/zesage/motion_sensors.git
-      ref: master
+  mobile_scanner: ^6.0.6
```
SputNikPlop, line 37: Not quite simple will break iOS builds

SputNikPlop:
As of release 2.10.2 everything has been included

PR #1334, opened 2025-02-26
fix: url package

SputNikPlop:
https://pub.dev/packages/flutter_custom_tabs/changelog to the latest version and opens possibility to use some of the new android features

### DIFF
--- a/lib/urls.dart
+++ b/lib/urls.dart
@@ -5,29 +5,27 @@ Future<void> openUrl(Uri url) async {
   if (!url.hasScheme) {
     await customtabs.launchUrl(
       url.replace(scheme: 'http'),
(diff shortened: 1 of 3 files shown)


kevmo314 (approved):
Looks good but seems like this might break the android build. I think that's actually addressed in https://github.com/muxable/rtchat/pull/1333 though, could you help clean up that PR to just updating the web auth package?

SputNikPlop:
> Looks good but seems like this might break the android build. I think that's actually addressed in #1333 though, could you help clean up that PR to just updating the web auth package?

That one is a mess, just pushed on this one a web auth package update

SputNikPlop:
@kevmo314 do you want me to switch the package to https://pub.dev/packages/flutter_web_auth_2/changelog

PR #1335, opened 2025-03-04
fix: event languages

SputNikPlop:
This is what I think you wanted as you described to me in the discord.


lib/components/channel_localization.dart
```diff
@@ -0,0 +1,43 @@
...
+    } else {
+      _channelLocale = null;
+    }
+    notifyListeners();
+  }
+
+  static Locale? getLocale(BuildContext context) {
+    return Provider.of<ChannelLocalization>(context, listen: false)
+        ._channelLocale;
+  }
```
kevmo314, line 19: This change looks good overall, but it would be simpler to put this function on `UserModel` and directly depend on it to avoid yet another change notifier.

SputNikPlop: Ok I think I fixed this

lib/components/channel_localization.dart
```diff
@@ -0,0 +1,43 @@
+import 'package:flutter/material.dart';
+import 'package:provider/provider.dart';
+
+class ChannelLocalization extends ChangeNotifier {
```
kevmo314, line 4: This entire class is not really necessary since you already have the property on `UserModel`. Listen on `UserModel` instead.

SputNikPlop:
@kevmo314 Anything else needed for this?

kevmo314 (approved):

PR #1339, opened 2025-04-17
fix: landscape commands

SputNikPlop:
Makes the commands appear in landscape without the long description kind of like emotes


kevmo314:
Seems reasonable, can you add a screenshot?

SputNikPlop:
> Seems reasonable, can you add a screenshot?

IMG_4346

kevmo314 (approved):

PR #1343, opened 2025-04-24
fix: use the new endpoint for ban, which is the same for timeout

chungwwei:
https://dev.twitch.tv/docs/api/reference/#ban-user


kevmo314 (approved):
Nice thanks for the fix and figuring that out!

PR #1356, opened 2025-05-25
fix: quality selector

SputNikPlop:


kevmo314:
Does this actually work? I recall when I implemented the preview, a lot of the resolutions had no effect on the quality of the stream that twitch sent even if the streamer had transcoding resources.

SputNikPlop:
> Does this actually work? I recall when I implemented the preview, a lot of the resolutions had no effect on the quality of the stream that twitch sent even if the streamer had transcoding resources.

Yes and no, works for partners but really not sure with enhanced broadcasting so lemme leave it open and test when I got time

kevmo314:
> > Does this actually work? I recall when I implemented the preview, a lot of the resolutions had no effect on the quality of the stream that twitch sent even if the streamer had transcoding resources.
> 
> Yes and no, works for partners but really not sure with enhanced broadcasting so lemme leave it open and test when I got time

Ok, the change otherwise seems reasonable. It would be nice if we can detect the list of functional resolutions and switch between those, otherwise the non-partners will end up with a button that does nothing most of the time.

SputNikPlop:
@kevmo314 Ok check the new changes out perhaps it works

assets/twitch-tunnel.js
```diff
@@ -37,6 +37,44 @@ window.action = function(eventName, params) {
...
+        }
+      } catch (e) {
+        console.error('Quality detection failed:', e);
+      }
+    }
+  }, 500);
+};
+
+// Initialize when Twitch player is ready
+if (typeof Twitch !== 'undefined') {
```
kevmo314: I don't think this exists?

image

SputNikPlop: Ok I tried some new js, let me know if that works bit of a patch work in me understanding whats going on

PR #1363, opened 2026-01-20
Add option to replace underscores in author names for TTS

LeMyst:
This pull request adds a new feature to the Text-to-Speech (TTS) system that allows users to replace underscores in author names with spaces, improving the readability of spoken names. The change includes UI, model, and persistence updates to support this new option.

**New feature: Underscore replacement in author names**

* Added `_isUnderscoreReplacementEnabled` property to `TtsModel`, with associated getter and setter, and included it in model serialization/deserialization to persist the setting. [[1]](diffhunk://#diff-8b4ba2729a4e3cc7350708b4ff34fc9a496e3f81e13484d125d36e2f78279acdR57) [[2]](diffhunk://#diff-8b4ba2729a4e3cc7350708b4ff34fc9a496e3f81e13484d125d36e2f78279acdR338-R346) [[3]](diffhunk://#diff-8b4ba2729a4e3cc7350708b4ff34fc9a496e3f81e13484d125d36e2f78279acdR528-R530) [[4]](diffhunk://#diff-8b4ba2729a4e3cc7350708b4ff34fc9a496e3f81e13484d125d36e2f78279acdR556)
* Modified the logic in `TtsModel` to replace underscores with spaces in author names when the new setting is enabled.

**User interface updates**

* Added a new switch in the TTS settings screen (`TextToSpeechScreen`) to allow users to toggle the underscore replacement feature.


kevmo314 (commented):
This seems reasonable but we should always replace underscores if it sounds better, it doesn't need to be a setting.

LeMyst:
> This seems reasonable but we should always replace underscores if it sounds better, it doesn't need to be a setting.

Maybe set it as true as default ? I don't know if there is many users that compained about this;

